PR #153, opened 2015-03-14
CSPACE-6585: Initial iteration of Ant buildfile to clone and update a ne...

aronr:
...w tenant folder in the Services layer.


aronr:
This still relies on a second (and third!) Ant buildfile, rather than the second Ant buildfile being integrated into the existing top-level Services buildfile. Closing pull request for now until I have a chance to discuss with @remillet.

PR #177, opened 2016-07-20
CSPACE-6947 and CSPACE-6948

lkvoong:
replaced SansSerif with Arial; updated query to display message when no objects are related.


lkvoong:
erroneously updated font.

PR #200, opened 2016-07-22
Add When No Data band; update query to allow for 0 records returned.

lkvoong:


lkvoong:
Message in NoData band does not specify "Loan Out".

PR #218, opened 2019-05-14
DRYD-568: Fixed issue when deleting term from a very long list.

remillet:
Fixed issue when deleting term from a very long list.


remillet:
See JIRA: https://issues.collectionspace.org/browse/DRYD-568

PR #224, opened 2019-08-31
Cc 740 5.2 - Support for salted passwords and tracking last login data.

remillet:
Changes for salting user passwords.  Added support for tracking "last login" for users.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ target
 m2-settings.xml
 *.log
 cspace-app-perflog.csv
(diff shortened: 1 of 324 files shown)


remillet:
For existing AuthN/AuthZ database:

{code}
ALTER TABLE users ADD COLUMN IF NOT EXISTS lastlogin timestamp without time zone;
ALTER TABLE users ADD COLUMN IF NOT EXISTS salt character varying(128);
{code}

PR #229, opened 2019-12-12
DRYD-477: Event handler refactor

remillet:
Refactored CSpace's event handling classes and interfaces to support both sync and async Nuxeo event handlers.


remillet:
All unit tests passed.

remillet:
NuxeoListenerRefactor

PR #237, opened 2020-04-17
DRYD-855: Change refname (/refobjs endpoint) searches to NXQL phrase searches rather than keyword searches

remillet:
Change refname (/refobjs endpoint) searches to NXQL phrase searches rather than keyword searches. SEE https://collectionspace.atlassian.net/browse/DRYD-855


remillet:
These changes have already been merged into the master branch.

PR #248, opened 2020-08-05
CC-1302: Service binding parameter values not handled correctly for N…

remillet:
CC-1302: Service binding parameter values not handled correctly for Nuxeo exent listeners.  Parameters are needed by a UCB listener to qualify different Nuxeo schemas where the "crate" location fields are stored.


ray-lee (commented):
Looks like this PR does a couple things in addition to what is stated. It also makes changes to the UpdateObjectLocationAndCrateOnMove handler, and registers that handler in Nuxeo.

I'm happy to merge a PR for what is stated. I need to think about the UpdateObjectLocationAndCrateOnMove stuff. Does it really need to be a separate handler at that point? If you're parameterizing it, could you add parameters to UpdateObjectLocationOnMove, to determine if it updates a crate field or not? Then we only need one handler.

remillet:
Since crates are not used by all tenants, I'd prefer to keep a separate handler/listener for dealing with crates.  UpdateObjectLocationOnMove is a subclass of UpdateObjectLocationOnMove.

However, if we moved support for crates into the common schema then having a single listener makes sense.  If and until we do that, I'd prefer to keep a separate (subclassed) listener for dealing with crate values.

PR #275, opened 2021-08-05
DRYD-921: fixed incorrect hierarchy.name for original nagpraReportFiledDate

lkvoong:
@ray-lee  please review.
cc @remillet


ray-lee (approved):
Looks good, thanks.

PR #295, opened 2022-05-13
DRYD-1036: IterationReport Procedure

mikejritter:
Jira Ticket: https://collectionspace.atlassian.net/browse/DRYD-1036

One thing to note - I opted to remove the `3rdparty` module as it's described as being obsolete and there for backwards compatibility. Since this is a new procedure I didn't see much of a reason to keep this around, but can revert the change if need be.


ray-lee:
Yup, it should be fine to omit `3rdparty`.

ray-lee (approved):
This looks good!

PR #297, opened 2022-06-14
DRYD-1036: Fix iteration report xsd

mikejritter:
* Use correct identifier in iterationreports_common xsd
* Update tests to use correct id


ray-lee (approved):
Looks good.

PR #301, opened 2022-08-30
Fcart es mapping update

mark-cooper:
Apply refname_displayname_normalizer to objectProductionPlace.
Field uses autocomplete on: place/local


ray-lee (approved):
Looks good, thanks.

PR #304, opened 2022-11-10
DRYD-1170: fix query in object_valuation report

kspurgin:
Just changes an erroneous INNER JOIN to a LEFT OUTER JOIN


mikejritter (approved):
Looks good. I tested using two objects, one with and one without a brief description, and was able to get both back from the query. Using the old query I only saw the object with the brief description set.

ray-lee (approved):

PR #307, opened 2022-12-14
DRYD-1183: Set totalItems and itemsInPage for derivative lists

mikejritter:
Jira: https://collectionspace.atlassian.net/browse/DRYD-1183

* Track number of derivatives for blobs when adding to the CommonList
* Minor code cleanup

## Notes

There isn't a test which covers the NuxeoBlobUtils, though I'm not against adding something to at least try and get something here to verify behavior. Otherwise to test what I did was:
* Build CSpace
* Create a media handling resource so that derivatives are generated
* Use curl to get the derivative list, e.g. `curl --user "admin@core.collectionspace.org:Administrator" http://localhost:8180/cspace-services/media/8f86f12b-ae09-4cf3-85d2/blob/derivatives | xmllint --format - | less`


ray-lee (approved):
Looks good to me.

PR #310, opened 2023-01-30
DRYD-1127: Use timestamp formatter for scalar values

mikejritter:
Jira: https://collectionspace.atlassian.net/browse/DRYD-1127

* Include timestamp when formatting structured date scalar values

# Notes

This is a breaking change, so this will be for 8.x. I'm keeping this a draft until that's ready.


mikejritter:
While working on some of the public browser tickets I noticed the `DefaultESDocumentWriter` subtracts a day from the latest scalar value, which seems like would be broken with these changes. Will need to look for other instances where the latest value is used and this could potentially break old behavior.

ray-lee (approved):
This looks good. I don't think the subtracting one that you saw is a problem. The latest scalar date has always been exclusive, and there's code all over the place to account for that, but from what I can see this doesn't change that. You removed `dateTime.setTime(23, 59, 59, 999);` on the latest scalar date, but I don't think that was the right thing to do in the first place, and in any case the time wasn't being printed. A date without time would have been interpreted as having time 00:00:00.000 when given back to the API, and this just makes that explicit.

PR #311, opened 2023-02-16
DRYD-1145: Get version info from implementation version

mikejritter:
Jira: https://collectionspace.atlassian.net/browse/DRYD-1145

* Add build info to manifest for jar and wars
* Use implementation version when getting systeminfo

# Notes

* I'm not sure if we should add something different for the build number. I was considering pulling in the git commit hash, but left it hardcoded at the moment.
* I added the build info to both war and jars. I wasn't quite sure how everything is bundled together and how the classpath is impacted, and this seemed like the easiest way to make sure it made it into the systeminfo modules.
* The diff was kind of all over the place with formatting initially so I had intellij reformat SystemInfoResource and made changes on top of that.

### DIFF
--- a/services/systeminfo/service/src/main/java/org/collectionspace/services/systeminfo/SystemInfoResource.java
+++ b/services/systeminfo/service/src/main/java/org/collectionspace/services/systeminfo/SystemInfoResource.java
@@ -28,6 +28,7 @@ import org.collectionspace.services.common.security.UnauthorizedException;
 @Consumes({"application/xml"})
 public class SystemInfoResource extends AbstractCollectionSpaceResourceImpl<SystemInfoCommon, SystemInfoCommon> {
 
+    private static final String VER_DISPLAY_NAME = "CollectionSpace Services v%s.%s";
     private static final int MAJOR_IDX = 0;
     private static final int MINOR_IDX = 1;
     private static final int PATCH_IDX = 2;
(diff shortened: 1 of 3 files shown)


services/systeminfo/service/src/main/java/org/collectionspace/services/systeminfo/SystemInfoResource.java
```diff
@@ -31,74 +28,87 @@
...
-    		}
-
-    	} catch(Exception e) {
-    		Response response = Response.status(Response.Status.BAD_REQUEST).entity(e.getMessage()).type("text/plain").build();
+        SystemInfoCommon result;
+
+        try {
+            result = new SystemInfoCommon();
+            result.setInstanceId(ServiceMain.getInstance().getCspaceInstanceId());
+            result.setDisplayName(ServiceMain.VER_DISPLAY_NAME);
```
ray-lee, line 63: `ServiceMain.VER_DISPLAY_NAME` currently contains a hardcoded version number. I think we should change that to be a template string, and fill in the major and minor numbers here with `String.format`.

ray-lee (changes requested):
This looks good so far. I think it would be good to set the build number to the short commit hash. Also, the display name needs to be updated, since it currently contains a hardcoded version number.

ray-lee (approved):
Looks good!

PR #317, opened 2023-02-24
DRYD-901: Only restrict fields for core namespace

mikejritter:
**What does this do?**

This updates the ExportResource to filter out restricted fields in the `collectionspace_core` namespace rather than the entire namespace. The set of allowed fields is now the `created{By,At}`, `updated{By,At}`, and `uri`. For namespaces which should have all fields removed, an empty list is provided to indicate the entire part should be removed.

In the export modal, it is possible to selected the created and updated fields for a record, but when getting the results these returned with empty values. In addition, it was requested to be able to export the uri in order to make automated operations easier.

* https://collectionspace.atlassian.net/browse/DRYD-901
* https://collectionspace.atlassian.net/browse/DRYD-1080 

Testing can be done through the export API to check that the requested fields from `collectionspace_core` can be returned.

* Create an Object record and take note of the csid
* Update the Object record so that it has updatedAt and updatedBy fields with data
* Create an xml payload for the invocation, replacing the objectCSID with the newly created one
```
<ns2:invocationContext xmlns:ns2="http://collectionspace.org/services/common/invocable">
        <mode>list</mode>
        <docType>CollectionObject</docType>
    <outputMIME>text/csv</outputMIME>
    <includeFields>
          <field>collectionspace_core:createdAt</field>
          <field>collectionspace_core:createdBy</field>
          <field>collectionspace_core:updatedAt</field>
          <field>collectionspace_core:updatedBy</field>
          <field>collectionspace_core:uri</field>
    </includeFields>
    <listCSIDs>
        <csid>${objectCSID}</csid>
    </listCSIDs>
</ns2:invocationContext>
```
* POST the xml to the export endpoint, e.g.
`curl -v -H "Content-Type: application/xml" -X POST --data @data.xml -u "admin@core.collectionspace.org:Administrator" http://localhost:8180/cspace-services/exports`
* Verify that the csv returned contains entries for the requested fields

None

There's a page for `Export Search Results` from the context of the ui which does not need to be updated. I'm not sure if there's a place for the endpoint which should be updated.

@mikejritter has tested on a local instance

### DIFF
--- a/services/export/service/src/main/java/org/collectionspace/services/export/ExportResource.java
+++ b/services/export/service/src/main/java/org/collectionspace/services/export/ExportResource.java
@@ -30,10 +30,11 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.HashSet;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 
+import com.google.common.collect.ImmutableMap;
 import org.apache.commons.lang3.StringUtils;
 import org.collectionspace.services.client.ExportClient;
 import org.collectionspace.services.client.PayloadOutputPart;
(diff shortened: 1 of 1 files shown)


ray-lee (approved):
This looks good!

PR #331, opened 2023-04-13
DRYD-1178: Create boxlist report

mikejritter:
**What does this do?**
Adds a `box_list` report

Report originally requested by OHC
https://collectionspace.atlassian.net/jira/software/c/projects/DRYD/issues/DRYD-1178

* Create a LMI with the following:
  * Current Location
  * Related Object w/ object name + related media  
* Register the report with the server for single docs:
```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<document name="report">
 <ns2:reports_common xmlns:ns2="http://collectionspace.org/services/report">
 <name>Box List</name>
 <notes>boxlist</notes>
 <forDocTypes>
   <forDocType>Movement</forDocType>
 </forDocTypes>
 <supportsSingleDoc>true</supportsSingleDoc>
 <supportsDocList>false</supportsDocList>
 <supportsGroup>false</supportsGroup>
 <supportsNoContext>true</supportsNoContext>
 <filename>box_list.jrxml</filename>
 <outputMIME>application/pdf</outputMIME>
 </ns2:reports_common>
</document>
```
* Navigate to the LMI and run the report
  * PDF output should display the thumbnail
  * CSV the thumbnail is omitted   

None

No

@mikejritter ran the report against an LMI with all fields present

### DIFF
--- a/services/report/3rdparty/jasper-cs-report/src/main/resources/box_list.jrxml
+++ b/services/report/3rdparty/jasper-cs-report/src/main/resources/box_list.jrxml
@@ -1,0 +1,201 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Created with Jaspersoft Studio version 6.20.1.final using JasperReports Library version 6.20.1-7584acb244139816654f64e2fd57a00d3e31921e  -->
+<jasperReport xmlns="http://jasperreports.sourceforge.net/jasperreports" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://jasperreports.sourceforge.net/jasperreports http://jasperreports.sourceforge.net/xsd/jasperreport.xsd" name="boxlist" pageWidth="842" pageHeight="595" orientation="Landscape" columnWidth="802" leftMargin="20" rightMargin="20" topMargin="20" bottomMargin="20" isIgnorePagination="true" uuid="38098e4f-3241-4565-bf21-09bee365eae1">
+	<property name="com.jaspersoft.studio.data.sql.tables" value=""/>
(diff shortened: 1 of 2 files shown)


ray-lee:
@mikejritter Could you start adding the XML payload as a .xml file alongside the .jrxml file? Give it the same name as the .jrxml (e.g. `box_list.xml`). I want to see if we can load these automatically when the services layer starts, so we don't have to do a separate step.

mikejritter:
@ray-lee I just pushed up `box_list.xml` for creating the report. I haven't been sure about what to set for all of the values (e.g. the `outputMIME` tag), but this is what I've been using for testing. The JIRAs have been kind of vague for single/group reports as well but pretty sure they're working as intended.

ray-lee:
Thanks. Let's make sure Jessi checks if the reports are created the way she wants when she looks at them on dev.

ray-lee (approved):
This looks fine to me. I wasn't sure how well the `WITH` queries would perform in large databases -- specifically, would this result in large temp tables with many rows that would be immediately discarded by the joins in the main query? But this seems to say that it will be fine, as long as you only reference each temp table once: https://www.postgresql.org/docs/current/queries-with.html#id-1.5.6.12.7

If you use this method, and need to reference a temp table more than once, you may want to consider using `NOT MATERIALIZED`, as described in that doc, if it seems appropriate.

ray-lee:
Actually the above is true only for Postgres 12 and above. So hosted instances will be fine, but if anyone is on an old version of Postgres and has a lot of records, it might be a problem. I would guess that no one is on Postgres < 12 and has a large enough database for it to matter, but who knows.

PR #332, opened 2023-04-14
DRYD-1176: Add xml payload for incomingloanletter

mikejritter:
**What does this do?**
Adds xml for registering the incoming loan letter report with cspace
Small edit to the incoming loan letter report to only fetch the first term group for person/organizations

To assist with registering reports

Test that the xml can be used in the report API call, e.g.
```
curl -X POST http://localhost:8180/cspace-services/reports -i -u admin@core.collectionspace.org:Administrator -H "Content-Type: application/xml" -T incoming_loan_letter.xml
``` 

n/a

n/a

@mikejritter previously used this for testing the incoming loan letter report


ray-lee (approved):
Thanks!

PR #339, opened 2023-04-24
DRYD-1195: Tombstone with budget + image report

mikejritter:
**What does this do?**
* Adds report for Tombstone with budget and related media thumbnail
* Adds xml for registering report

Jira: https://collectionspace.atlassian.net/browse/DRYD-1195

This is a report for publicart (I assume requested by them though the jira does not state this)

* Create the report for the publicart tenant w/ the provided xml
* Create a CollectionObject record with the following fields:
  * Title, Creator, Brief Description, Collection, Owner, Installation Type, Work Type, Material, Dimension, Artwork Date
* Create a related Acquisition with the follow fields:
  * Funding value, funding currency
* Create a related Media with an image for the thumbnail
* From the search page, select the CollectionObject and run the report 

The jira states that the Artwork Creators should be limited to `artist` and `lead artist`, however I only see `artist` defined in the base vocabularies. I can update the query to include what _might_ be the short id of the lead but I left it with only artist for now.

Additionally the Creators field can join on more than one row, and I'm not sure if there's a good way to display this so I opted to keep it simple and just return multiple rows. I was initially going to try and group by the artwork id, but since this is supposed to be csv we need a delimiter + a way to deref multiple vals (unless that already works).

No

@mikejritter Ran with all fields existing, and multiple creators of different types

### DIFF
--- a/services/report/3rdparty/jasper-cs-report/src/main/resources/tombstone_with_budget.xml
+++ b/services/report/3rdparty/jasper-cs-report/src/main/resources/tombstone_with_budget.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<document name="report">
+  <ns2:reports_common xmlns:ns2="http://collectionspace.org/services/report">
+    <name>Tombstone with Budget</name>
+    <notes>Tombstone Report with Image and Budget (Public Art Profile)</notes>
+    <forDocTypes>
+      <forDocType>CollectionObject</forDocType>
+    </forDocTypes>
+    <supportsSingleDoc>false</supportsSingleDoc>
+    <supportsDocList>true</supportsDocList>
+    <supportsGroup>false</supportsGroup>
+    <supportsNoContext>true</supportsNoContext>
+    <filename>tombstone_with_budget.jrxml</filename>
+    <outputMIME>application/pdf</outputMIME>
+  </ns2:reports_common>
+</document>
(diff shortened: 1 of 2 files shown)


services/report/3rdparty/jasper-cs-report/src/main/resources/tombstone_with_budget.xml
```diff
@@ -0,0 +1,16 @@
...
+    <notes>Tombstone Report with Image and Budget (Public Art Profile)</notes>
+    <forDocTypes>
+      <forDocType>CollectionObject</forDocType>
+    </forDocTypes>
+    <supportsSingleDoc>false</supportsSingleDoc>
+    <supportsDocList>true</supportsDocList>
+    <supportsGroup>false</supportsGroup>
+    <supportsNoContext>true</supportsNoContext>
+    <filename>tombstone_with_budget.jrxml</filename>
+    <outputMIME>application/pdf</outputMIME>
```
ray-lee, line 14: If this is intended to be a csv, you should set `outputMIME` to `text/csv`. That will make CSV the default format when the report is run.

ray-lee (commented):
If we decide on what the delimiter should be for creators, we can probably have the deurning function handle it. Another way might be to select the creators into an array of strings. Then I think you can write a subreport that formats the array, where you might be able to just use the current way of deurning, but I've never tried this, and don't exactly know how it works in Jasper. It might be overkill for this.

services/report/3rdparty/jasper-cs-report/src/main/resources/tombstone_with_budget.xml
mikejritter: Yea I had been using pdf for testing so I could see the thumbnails but can switch things over to start using csv.

ray-lee: That's a good point... I'm not sure what the intent is, if it's asking for an image in a csv-formatted report. We should probably clarify, but this looks good for now.

PR #340, opened 2023-04-26
DRYD-1197: Object with Full Current Place Report

mikejritter:
**What does this do?**
* Creates report for object w/ current place

Jira: https://collectionspace.atlassian.net/browse/DRYD-1197
Report for publicart

* Register the report with collectionspace using the provided xml
  * title, description
* Create a place authority with
  * type, owner (all), address (all), latitude, longitude, and broader place
* Create a movement with
  * current location as the above authority, current location note
* Select one or more of the search results and run the report

This is for the publicart profile, so we just need to make sure the reported is only loaded on that profile 

No

@mikejritter ran the report with all fields present


ray-lee (approved):
Looks good.

PR #341, opened 2023-04-26
DRYD-1199: Basic Object with Computed Location Report

mikejritter:
**What does this do?**
* Adds report for Object w/ Computed Current Location
* Adds xml for register report

Jira: https://collectionspace.atlassian.net/browse/DRYD-1199
Report originally requested by Public Art

* Register the report with cspace
  * title, object name, brief description
* Create a location authority with
  * term display name and term name
* Create a LMI with the above authority as the current location
* Relate the LMI and collectionobject
* Create a related media for the collectionobject
* Search on collectionobjects and execute the report

The Jira states `Storage Location or Place name` for one of the items, however on core this is from the `location` and `organization` authorities. Though I think it would be a good idea to add a few joins to get `place` if that exists so it can be complete on other tenants.

There was also a request to get the taxon name when both the title and object name are not present. I added this though it feels a little clunky as it uses the column for the object name. It almost seems better to create a separate report with the taxon available for profiles which use it imo, but might be worth discussion more.

No

@mikejritter tested on core and anthro


ray-lee (approved):
Looks good.

PR #342, opened 2023-04-27
Update relation queries for reports

mikejritter:
**What does this do?**
This PR updates many of the queries for the new csv reports in order to ignore relations on objects/procedures which have been deleted. 

When testing on dev we saw some reports generating many more rows than expected and when looking into the cause I found that there were deleted relations being retrieved in the report. This work is being done in order to remove those rows from the reports.

I also updated the default output to be `text/csv`.

Most of the reports only query for related objects and media on said objects, what I have is:
* A single base collectionobject and 2 media handling procedures
* Relate the collectionobject with the media
* Delete on of the media procedures

Then for run reports which query related collectionobjects/media
* Accessions
  * Acquisition -> CollectionObject -> Media
*  Box List
    * LMI -> CollectionObject -> Media
* Deaccessions
  * ObjectExit -> CollectionObject -> Media
  * Also needs test on related acquisitions and intake
* Deed of Gift
  * Intake -> CollectionObject -> Media
* Object w/ Computed location
  * CollectionObject -> Media
* Tombstone w/ budget
  * CollectionObject -> Media
  * Also needs test on related acqusition

I updated each query to be consistent with how I had previously done them. So if the report already used  CTEs I went that route, otherwise I found I could use subqueries and join on them. I'm not really sure about performance implications for them, but both allow an easy way to make sure we 1) get a row back if no relation exists and 2) avoid any records which have been deleted. 

I also only updated the sql in the jasper reports, and if needed I can reimport the reports into jasper to regenerate the `<field>` tags. It didn't seem like this impacted the ability of the server to run the reports but I'm still going through them.

No

@mikejritter tested the accessions and object w/ computed location reports. Other queries have been ran but I reset my database the other day so I have no test data; still going through the rest of the reports to verify.

### DIFF
--- a/services/report/3rdparty/jasper-cs-report/src/main/resources/accessions.jrxml
+++ b/services/report/3rdparty/jasper-cs-report/src/main/resources/accessions.jrxml
@@ -56,24 +56,31 @@
 	$P!{whereclause}
 ), related_objects AS (
   SELECT
+    hier.name AS csid,
     object.collection,
     object.objecthistorynote,
     object.objectnumber,
     ong.objectname,
     bd.item AS briefdescription,
     sdg.datedisplaydate AS objectproductiondate,
-    media.objectcsid AS mediacsid,
     acq.csid AS acquisitioncsid
   FROM acquisitions acq
   INNER JOIN relations_common rels ON rels.subjectcsid = acq.csid AND rels.objectdocumenttype = 'CollectionObject'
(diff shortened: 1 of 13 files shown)


mikejritter:
I just did some additional testing, the `tombstone_with_budget` report has an error in it as I got a 500 back when running it. Should be fixed shortly.

ray-lee (approved):
I think the `<field>` tags only matter if you add a new column, and need to reference it in the report. It doesn't look like that happened. As long as the reports run, it should be fine.

PR #350, opened 2023-05-24
DRYD-1154: De-urn reports

mikejritter:
**What does this do?**
Updates reports which have refnames to use the new deurnfields parameter

Jira: https://collectionspace.atlassian.net/browse/DRYD-1154

Many of the older reports needed to be updated in order to have any refnames de-urned. This is done by using the deurnfields parameter to handle refnames which appear in the details section of the reports.

Testing requires running of each report in order to validate that the fields returned are de-urned. Many reports retrieve a related collectionobject, which should be created.
* Replace the reports with the updated jrxml
* Create a collectionobject with an objectname, production person role, production person, and computed location
* For each updated report, create a procedure and relate the collectionobject, then run the report
  - [x] Acquisition basic
  - [x] Acquisition Ethnographic
  - [x] Condition check basic
  - [x] Exhibition basic
  - [x] Group Obj Ethnographic  
  - [x] Intake Ethnographic
  - [x] LoanIn Basic
  - [x] LoanIn Ethnographic
  - [x] LoanOut Basic
  - [x] LoanOut Ethnographic
  - [x] ObjectExit Ethnographic 

The commits for each individual jira, and would likely be best to merged without a squash.

If a report has already been run, it will need to be recompiled by collectionspace before the updates are visible. Just something to look out for when testing, especially on the dev deployment.

No

@mikejritter tested each checked report

### DIFF
--- a/services/report/3rdparty/jasper-cs-report/src/main/resources/coreIntake.jrxml
+++ b/services/report/3rdparty/jasper-cs-report/src/main/resources/coreIntake.jrxml
@@ -7,6 +7,9 @@
 	<style name="SubTitle" forecolor="#666666" fontName="SansSerif" fontSize="18"/>
 	<style name="Column header" forecolor="#666666" fontName="SansSerif" fontSize="12" isBold="true"/>
 	<style name="Detail" fontName="SansSerif" fontSize="12"/>
+	<parameter name="deurnfields" class="java.lang.String" isForPrompting="false">
+		<defaultValueExpression><![CDATA["site,objectname,productionpeople"]]></defaultValueExpression>
+	</parameter>
 	<parameter name="tenantid" class="java.lang.String" isForPrompting="false">
 		<defaultValueExpression><![CDATA[15]]></defaultValueExpression>
 	</parameter>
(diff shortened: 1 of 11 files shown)


services/report/3rdparty/jasper-cs-report/src/main/resources/coreIntake.jrxml
```diff
@@ -21,19 +24,13 @@ inc.entryDate entryDate,
...
-case when (co.fieldcollectionplace  is not null and co.fieldcollectionplace  <> '') then
- regexp_replace(co.fieldcollectionplace , '^.*\)''(.*)''$', '\1')
- else ''
-end AS site,
+co.fieldcollectionplace AS site,
 ong.objectName AS "objectname",
 case when (bd.item is not null and bd.item <> '') then
  regexp_replace(bd.item, '^.*\)''(.*)''$', '\1')
  else ''
 end AS description,
```
ray-lee, line 32: @mikejritter Do you know what's going on with this description field? It's still being de-urned in SQL. I suspect the reports that do this came from Berkeley, and they probably have some museum that uses this field to store a ref name.

ray-lee: I looked through the Berkeley configs and I don't see anywhere where they changed briefDescription to hold a ref name. So I think we can get rid of this? It appears in a couple other reports too.

mikejritter: I went ahead and removed the de-urning for the briefDescription fields. 

I had also tried on some of the remaining fields which use the regexp_replace logic (e.g. donor in coreAcquisition) but they weren't being de-urned when running the report. They all exist in the `<title>` section, and I took a brief look at the jasper scriplet API but didn't see any methods for that part of the lifecycle. Doesn't seem like a big deal atm since it's still being handled in the query.

ray-lee: That sounds fine, thanks!

ray-lee (approved):

PR #351, opened 2023-05-25
Update jasper and reformat reports

mikejritter:
**What does this do?**
* Updates jasper to 6.11.0
* Updates formatting for LoanOut Basic List
* Updates formatting for Group Object Ethnographic List 

Jira: https://collectionspace.atlassian.net/browse/DRYD-1164, https://collectionspace.atlassian.net/browse/DRYD-1159

Formatting changes were requested in various jira issues in addition to de-urning. As part of this, there's a new attribute for `textField` so that different approaches can be taken depending on the report. Jasper is being updated as well in order to make use of the new attribute. When nuxeo is updated further we'll have an opportunity to bring Jasper fully up to date.

* Rebuild collectionspace 
* Run the Group Object Ethnographic List with a long title 
  * This should scale the font size down in order to avoid overlapping with the owner below
  * `StretchHeight` would be nice to use here, but it wasn't having any effect when testing. I'm not sure if it was the layout impacting it or what, but that's the reason for using `ScaleFont`
* Run the LoanOut Basic List and check the dimension and number of objects columns 

None

No

@mikejritter tested against the two updated reports

### DIFF
--- a/services/report/3rdparty/jasper-cs-report/src/main/resources/LoansOut_List_Basic.jrxml
+++ b/services/report/3rdparty/jasper-cs-report/src/main/resources/LoansOut_List_Basic.jrxml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- Created with Jaspersoft Studio version 6.1.1.final using JasperReports Library version 6.1.1  -->
 <!-- 2015-12-16T19:37:02 -->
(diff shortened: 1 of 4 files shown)


services/report/3rdparty/jasper-cs-report/src/main/resources/LoansOut_List_Basic.jrxml
```diff
@@ -207,7 +207,7 @@
 	<noData>
 		<band height="50">
 			<staticText>
-				<reportElement uuid="f449dbd9-0782-4ffc-9296-1c09a978c94b" style="Column header" x="0" y="22" width="375" height="28"/>
+				<reportElement uuid="f449dbd9-0782-4ffc-9296-1c09a978c94b" style="Column header" x="1300" y="22" width="375" height="28"/>
```
ray-lee, line 210: `x="1300"` is pushing this text off the page. Otherwise, this looks good.

ray-lee (approved):

PR #352, opened 2023-06-05
Set overflow and null rules for results

mikejritter:
**What does this do?**
* Set overflow and null rules for results

This issue was brought up in [DRYD-1198](https://collectionspace.atlassian.net/browse/DRYD-1198) as results were being truncated in the csv report for certain fields. I decided to copy some of the older reports which set overflow rules (using `textAdjust`) and apply them to all fields in the result set so that none of them will have truncation applied and will instead "stretch". In addition I updated the reports so that null fields will be blank instead of returning `null`, which had been done in other reports as well.

* For one of the given reports, create a procedure or object that has enough data to cause an overflow in one of its fields (I'm not sure what the previous limit was it seems like it's ~21 or 22 characters)
* Run the report and see that the returned data is not truncated
* Since the change is the same for each field, testing all shouldn't be necessary but would be ideal 

None

No

@mikejritter tested on the condition report


ray-lee (approved):
Looks good.

PR #354, opened 2023-06-13
Add combine param to report and batch searches.

ray-lee:
**What does this do?**

This adds a `combine` parameter to list requests for report and batch resources. This controls how the filters (`doctype`, `filename`, `classname`, `mode`) are combined. Set to `or` to combine the filters using an OR operation. No value or any other value will combine the filters using and AND operation.

When viewing a Group record, the reports and batch jobs shown in the sidebar should be _either_ registered for doc type `Group`, _or_ for invocation mode `group`. There was previously no way to combine these filters using an OR operator.

JIRA: https://collectionspace.atlassian.net/browse/DRYD-1069

With the default reports installed:
- Request `/cspace-services/reports?doctype=Group`
   There should be one result, Group Object Ethnographic Object List.
- Request `/cspace-services/reports?mode=group`
   There should be one result, Group Basic List.
- Request `/cspace-services/reports?doctype=Group&mode=group`
   There should be no results.
- Request `/cspace-services/reports?doctype=Group&mode=group&combine=and`
   There should be no results.
- Request `/cspace-services/reports?doctype=Group&mode=group&combine=or`
   There should be two results (both of the above).

None.

I will update the REST API documentation at https://collectionspace.atlassian.net/wiki/spaces/UNR/pages/1565362629/Batch+Job+Service+RESTful+APIs and https://collectionspace.atlassian.net/wiki/spaces/UNR/pages/1565363217/Reporting+Service+RESTful+APIs

@ray-lee ran this locally.

### DIFF
--- a/services/batch/service/src/main/java/org/collectionspace/services/batch/BatchResource.java
+++ b/services/batch/service/src/main/java/org/collectionspace/services/batch/BatchResource.java
@@ -99,6 +99,12 @@ public class BatchResource extends NuxeoBasedResource {
             String docType = queryParams.getFirst(IQueryManager.SEARCH_TYPE_DOCTYPE);
             String className = queryParams.getFirst(IQueryManager.SEARCH_TYPE_CLASS_NAME);
             List<String> modes = queryParams.get(IQueryManager.SEARCH_TYPE_INVOCATION_MODE);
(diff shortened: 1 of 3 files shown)


ray-lee:
@mikejritter FYI

ray-lee:
@mikejritter Actually I just realized this won't work, because we need to do (doctype=Group AND mode=single) OR (mode = group). So it's not just a matter or OR'ing all the filters.

PR #357, opened 2023-06-22
Aggregation for repeating fields in reports

mikejritter:
**What does this do?**
This PR updates many of the report queries to perform aggregation on certain fields which exist either in groups or simple lists. Field separators are done with a pipe, and de-urning of fields is done in the query using a regular expression.

Summary of changes are:
* DRYD-1118 (deedofgift): Aggregate termdisplayname
* DRYD-1129 (accessions): Aggregate approvalGroup
* DRYD-1130 (deaccessions): Aggregate deacApprovalGroup; fix typo for deaccession approval note
* DRYD-1132 (referral): Aggregate approvalGroup
* DRYD-1195 (tombstone with budget): Aggregate all fields
* DRYD-1196 (tombstone with creator): Aggregate creator, creator role, material, and objectname

These were requested from feedback as something which is desired in the reports. The fields chosen to aggregate are limited to what was requested in each JIRA ticket.

* Redeploy the updated reports
* For each report, run for the given procedure type to ensure that the report compiles and completes successfully

I want to double check the list of changes w/ Jessi to make sure we have everything that's requested.

@mikejritter ran each updated report


ray-lee (approved):
Looks good.

PR #358, opened 2023-06-26
Additional Report Updates

mikejritter:
**What does this do?**
Brings in additional changes requested for reports:

* DRYD-1118: Ensure ordering for term names
* DRYD-1130: Add exit quantity
* DRYD-1196: Aggregate dimension values

These changes are the latest requested from testing the reports. For deed of gift, we want to make sure the primary change is first, so we use order by on the aggregation. Deaccession had a typo in the original requested changes which resulted in exit quantity being removed; this has been re-added. In tombstone with creator, the dimension fields were requested to be aggregated. 

* https://collectionspace.atlassian.net/browse/DRYD-1118
* https://collectionspace.atlassian.net/browse/DRYD-1130
* https://collectionspace.atlassian.net/browse/DRYD-1196

* Redeploy the reports
* Run the deed of gift report on an Intake procedure; see the correct order for the owner term names
* Run the deaccession report for Object Exits; see the exit quantity field returned
* Run the tombstone with creator report for CollectionObjects; see aggregated dimensions

None

No

@mikejritter tested locally

### DIFF
--- a/services/report/3rdparty/jasper-cs-report/src/main/resources/deaccessions.jrxml
+++ b/services/report/3rdparty/jasper-cs-report/src/main/resources/deaccessions.jrxml
@@ -26,6 +26,7 @@
     exit.id AS id,
     exit.exitnote,
     exit.exitnumber,
+    exit.exitquantity,
     exit.exitreason,
     exit.packingnote,
     exit.deaccessiondate,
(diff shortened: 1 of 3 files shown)


ray-lee (approved):
Looks good.

PR #360, opened 2023-07-25
Publicart Report Updates

mikejritter:
**What does this do?**
* Limits the thumbnails to a single row based on the most recently updated
* Aggregates artwork date in tombstone with creator report

Jira: https://collectionspace.atlassian.net/browse/DRYD-1195
Jira: https://collectionspace.atlassian.net/browse/DRYD-1196

Changes were requested as part of QA testing.

* Deploy the new jrxml into the `tomcat/cspace/reports` dir
* Create a CollectionObject with 2 relating Media handling procedures, multiple production dates
* Search for CollectionObjects and run the 2 reports (with output set to pdf): Tombstone with Budget; Tombstone with Creator
* For each report verify:
  * There is only a single thumbnail per collectionobject
  * The artwork date is aggregated in both

None

No


ray-lee (approved):
Looks good.

PR #361, opened 2023-08-30
DRYD-1248: UCB - Duplicate Object Number Report

mikejritter:
**What does this do?**
* Adds report for detecting duplicate object numbers to the botgarden profile

Jira: https://collectionspace.atlassian.net/jira/software/c/projects/DRYD/issues/DRYD-1248

This is part of the UCB bundle and is a report which can easily be added without needing other parts of the bundle. It appears to be for botgarden (see the join on `collectionobjects_naturalhistory`), so I've opted to only register it in the botgarden tenant bindings.

Changes to the original inclusion in the bundle are fairly minor, I believe I just ran the report through an xml formatter and updated some of the overflow rules to use the `textAdjust` attribute.

* Build collectionspace with the botgarden tenant enabled
* Verify the `Duplicate Object Report` can be seen when searching for collectionobjects
* Verify that the `Duplicate Object Report` can be run

Additional changes might be made to the registration payload -- we should verify how the report should be run so we know which `supports` flags to use. @remillet might be able to give some insight on this point.

There are some updates we can make to the report as well that I didn't do:
* Add tenantId parameter and add additional clause for core.tenantid = tenantid
  * This would bring the report more in line with the other reports, though not a major issue 
* Use the `deurnfields` parameter on `tig.taxon` (determination) and `fc.item` (collector)
  * Once again not major, just helps to clean up the query a little

No

@mikejritter has only verified that the sql and report execute, needs to run against dataset with duplicate object numbers

### DIFF
--- a/services/common/src/main/cspace/config/services/tenants/botgarden/botgarden-tenant-bindings.delta.xml
+++ b/services/common/src/main/cspace/config/services/tenants/botgarden/botgarden-tenant-bindings.delta.xml
@@ -44,5 +44,13 @@
             </service:initHandler>
         </tenant:serviceBindings>
 
+        <tenant:serviceBindings merge:matcher="id" id="Reports">
+            <service:properties xmlns:service="http://collectionspace.org/services/config/service" xmlns:types="http://collectionspace.org/services/config/types">
+                <types:item merge:matcher="skip" merge:action="insert">
+                    <types:key>report</types:key>
+                    <types:value>duplicateobjectnumber</types:value>
+                </types:item>
+            </service:properties>
+        </tenant:serviceBindings>
     </tenant:tenantBinding>
 </tenant:TenantBindingConfig>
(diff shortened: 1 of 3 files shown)


remillet (approved):
Perhaps we should make the 'duplicateobjectnumber' report be a UCB-only report.  I see the current UCB branch has the source file here: ./services/report/3rdparty/jasper-cs-report/src/main/resources/tenants/ucjeps/duplicateobjectnumber.jrxml

kspurgin:
I'd propose that adding a duplicate object number report as a core report would be a good idea, in terms of adding more support for maintaining data quality/integrity in CollectionSpace across the board. 

The CSV Importer can only update Object records that have unique object numbers, but there's no way for users to identify objects with non-unique numbers.

Not sure if there is something UCB specific about this particular report, but I'm in favor of a report for this purpose in general.

ray-lee:
Since this is botgarden-specific, and maybe even UCBG-specific (the "Determination" column doesn't work for a standard botgarden tenant), I'm going to close this without merging. We can make a ticket for a generic duplicate object number report.

PR #362, opened 2023-09-01
DRYD-1253: UCB Botgarden Contributions

mikejritter:
**What does this do?**
Adds contributions to the botgarden tenant from UCB

* Create tenant module for event listeners
* Move botgarden event listeners into their own botgarden module 
* Add doShortIdsMatch for comparing refnames
* Migrate constants to TaxonNaturalHistoryConstants
* Migrate constants to CollectionObjectNaturalHistoryConstants
* Create UpdateHybridNameBatchJob
* Create UpdatePlantPhrenologyBatchJob
* Create Pahma and UCJeps ValidationHandlers

**Additional Updates**
I made some minor qol changes to the contributions, including:
* Update to use parameterized logging
* Use charAt over substring when getting a single character
* Use Collections.singletonList instead of Arrays.asList
* Use equals instead of ==

I also updated the wiki link included in one of the commits, however I think with every update to the wiki the url will need to be updated. It might be better to just remove it and only mention the page title, if necessary.

Jira: https://collectionspace.atlassian.net/browse/DRYD-1253

This is part of the UCB contributions and brings in refactoring changes made to the botgarden tenant. In addition there are changes made to the batch jobs and listeners to check refnames in a more idiomatic way rather than just string equality.

The most straightforward way to test these changes is running the integration tests.
* Build and deploy collectionspace
* Use `mvn test` to run the integration test suite

The batch job and listener updates should be tested as well, though those would have to be done individually. 

I don't _think_ this will have any impact on releases. I was able to build with the botgarden tenant enabled and run against the ui without issue.

No

@mikejritter tested compilation, deployment, and integration tests for these changes


mikejritter:
@ray-lee I put this up as a draft so you can see the refactor changes. Mostly just the additional `tenant` module in `3rdparty/nuxeo/nuxeo-platform-listener`, and pushing `botgarden` into that. The other refactoring changes are moving constants around, which aren't really significant. I _think_ this goes along with some other changes which allow per-tenant sql during init, but neither are dependent on the other.

remillet (approved):
These changes seem fine.

services/batch/service/src/main/java/org/collectionspace/services/batch/nuxeo/botgarden/UpdatePlantPhenologyBatchJob.java
```diff
@@ -0,0 +1,135 @@
...
+        return results;
+    }
+
+    public void updateRecord(String csid, HashMap<String, String> values) throws URISyntaxException {
+        StringBuilder valuesToUpdate = new StringBuilder();
+        String sex = "";
+
+        for (String key : values.keySet()) {
+            if (key.equals("sex")) {
+                sex = "<sex>" + values.get(key) + "</sex>";
```
ray-lee, line 93: The value needs to be XML-escaped.

services/batch/service/src/main/java/org/collectionspace/services/batch/nuxeo/botgarden/UpdatePlantPhenologyBatchJob.java
```diff
@@ -0,0 +1,135 @@
...
+            } else {
+                valuesToUpdate.append("<").append(key).append(">")
+                              .append(values.get(key))
```
ray-lee, line 96: The value needs to be XML-escaped.

services/batch/service/src/main/java/org/collectionspace/services/batch/nuxeo/botgarden/UpdatePlantPhenologyBatchJob.java
```diff
@@ -0,0 +1,135 @@
...
+import org.collectionspace.services.common.invocable.InvocationContext;
+import org.collectionspace.services.common.invocable.InvocationResults;
+import org.collectionspace.services.common.invocable.InvocationContext.Params.Param;
+import org.collectionspace.services.common.ResourceMap;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This batch job updates the fruiting and flowering months, as well as the sex, fruit color and flower color of collectionobject records.
```
ray-lee, line 21: Where are these fields defined? I know where `sex` is. I expected to see the rest in botgarden-collectionobject.xm, but I'm not seeing them: https://github.com/collectionspace/application/blob/master/tomcat-main/src/main/resources/tenants/botgarden/botgarden-collectionobject.xml

Contrary to this comment, this appears to allow updating _any_ field in the `collectionobjects_botgarden` schema (along with `sex` in `collectionobjects_common`). @remillet is that the intent? If so, the name of the batch job and the above comment should probably change to make it more clear.

ray-lee (changes requested):
This mostly looks fine, but I had some questions about one of the new batch jobs, and noticed a couple other minor things.

services/batch/service/src/main/java/org/collectionspace/services/batch/nuxeo/botgarden/UpdatePlantPhenologyBatchJob.java
ray-lee: @remillet @rjaffe Also it would be good to know how this is used. Is there a UI? Or is this meant to be called by scripts?

rjaffe: @ray-lee UCBG has a Update Plant Phenology data update tool in the UI. See attached screenshots. Happy to discuss further. - Rick

ucbg qa collectionspace-Update-Plant-Phenology-data-update-tool-ui
ucbg qa collectionspace-data-update-tool-list

ray-lee: Ok, I'm fine with this as long as we rename it to something descriptive of what it actually does, and change its comment to describe to what actually it does.

ray-lee (approved):
Looks good now.

PR #370, opened 2023-10-03
Workflow Reports and Report Updates

mikejritter:
**What does this do?**
This brings in new reports as well as fixes to reports from 7.2:
* Add Borrowing Receipt report to core
* Add Artwork Description report to fcart
* Add Full Object With Place Details report to publicart
* Add PlaceType field to Object with Full Current Place
* Remove condition on primary type for relation joins
* Deurn objectname field

**New Reports**
Jira: https://collectionspace.atlassian.net/browse/DRYD-1278

These are the 3 new reports which were requested. Fairly straightforward additions. 

**Report Updates**
Jira: https://collectionspace.atlassian.net/browse/DRYD-1282

The updates are from things we saw after 7.2 release which weren't caught by QA. Notably, the `placeType` field was missing from the object w/ current place report. 

It was reported that related objects weren't being returned for a few reports (accessions, deed of gift, box list). This was being caused by the primary type having a different name for tenants, so I opted to drop the condition from the join since we're already using the csid. 

Object names weren't being deurned when running under publicart, and I saw they override the field to be controlled. This was resolved by adding `objectName` to the deurn fields list.  

* Enable the core, fcart, and publicart tenants
* Deploy collectionspace
* Test the borrowing receipt runs for core
  * From loan out procedure. Relies on a related collectionobject with a location and media handling.
  * See https://collectionspace.atlassian.net/browse/DRYD-1279 for all fields
* Test the artwork description report runs for fcart
  * From collection object
  * See https://collectionspace.atlassian.net/browse/DRYD-1280 for all fields
* Test the Full Object with Place Details runs for publicart
  * From collection object with related lmi and place authority
  * See https://collectionspace.atlassian.net/browse/DRYD-1281 for all fields
* Updated reports
  * object with full place should have the place type ~~, and deurn objectname if it's controlled~~
  * accessions should deurn objectname if it's controlled and join all related collection objects for non-core tenants
  * deed of gift should deurn objectname if it's controlled and join all related collection objects for non-core tenants
  * box list should join all related collection objects for non-core tenants
  * deaccessions should deurn objectname if it's controlled
  * incoming loan should deurn object name if it's controlled
  * outgoing loan should deurn object name if it's controlled
  * referral should deurn object name if it's controlled

I had been working off of the 7.2 release tag because I didn't feel like deploying the SSO changes in my local install yet. Can rebase if that makes testing easier now since it looks like everything is ready.

I still have an open question about the object name field for the new reports, which should be resolved shortly.

No

@mikejritter tested the new reports and updates

### DIFF
--- a/services/common/src/main/cspace/config/services/tenants/fcart/fcart-tenant-bindings.delta.xml
+++ b/services/common/src/main/cspace/config/services/tenants/fcart/fcart-tenant-bindings.delta.xml
@@ -270,5 +270,14 @@
 				}
 			</tenant:mapping>
 		</tenant:elasticSearchIndexConfig>
+
+		<tenant:serviceBindings merge:matcher="id" id="Reports">
+			<service:properties xmlns:service="http://collectionspace.org/services/config/service" xmlns:types="http://collectionspace.org/services/config/types">
+				<types:item merge:matcher="skip" merge:action="insert">
+					<types:key>report</types:key>
+					<types:value>artwork_description</types:value>
+				</types:item>
+			</service:properties>
+		</tenant:serviceBindings>
 	</tenant:tenantBinding>
 </tenant:TenantBindingConfig>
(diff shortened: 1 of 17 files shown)


ray-lee (approved):
Looks good.

PR #371, opened 2023-10-06
DRYD-1254: UCB Per Tenant SQL

mikejritter:
**What does this do?**
* Allow for per tenant sql initialization
* Remove support for MySQL db creation

**Additional Changes**
* Switch to try w/ resources when executing sql
* Adjust logging to use parameterized calls
* Remove some log level checks
* Minor formatting/whitespace updates
* Remove shortName from tenant config as it was added from SSO work

This is part of the UCB contributions and allows for tenants to have their own init scripts separate from the common scripts, presumably used by botgarden though those scripts were not included.

Run the Integration Tests:
* Run `mvn test` to ensure all tests pass

This adds a new attribute `shortName` to the tenant config. This might be considered a breaking change.

The runsqlscripts service binding is setup to be run for all tenants, including a script `grant_reporters_privileges.sql` which was not included. Likely best option is to remove the script. I left a comment on the change as well. 

Another script `create_unaccent_text_search_configuration.sql` was included which creates an extension `unaccent`. I don't know if this is something we would want, but isn't used. I thought it could use some discussion before removing from the changeset.

Finally the changes to ServiceMain also create a `utils` schema in postgres. Once again this isn't something that we make use of, so it might be worth omitting for now. I've left it in for more discussion, similar to the above.

No

@mikejritter verified integration tests run and cspace runs


services/common/src/main/cspace/config/services/tenants/tenant-bindings-proto-unified.xml
```diff
@@ -63,6 +63,23 @@
...
+			<service:initHandler xmlns:service="http://collectionspace.org/services/config/service">
+				<service:classname>org.collectionspace.services.common.init.RunSqlScripts</service:classname>
+				<service:params>
+					<service:property>
+						<service:key>sqlScriptName</service:key>
+						<service:value>grant_reporters_privileges.sql</service:value>
+					</service:property>
+				</service:params>
+			</service:initHandler>
+		</tenant:serviceBindings>
```
mikejritter, line 80: This script wasn't included, likely worth removing and just having an empty declaration which tenants can override.

services/common/src/main/resources/db/postgresql/create_unaccent_text_search_configuration.sql
mikejritter, line 1: This file was added but isn't used. Likely we can just remove it.

services/common/src/main/java/org/collectionspace/services/common/ServiceMain.java
```diff
@@ -974,7 +1011,8 @@ private HashSet<String> createNuxeoDatabases() throws Exception {
 							JDBCTools.createNewDatabaseUser(JDBCTools.CSADMIN_DATASOURCE_NAME, repositoryName, cspaceInstanceId, dbType, readerUser, readerPW);
 						}
 						// Create the database
-						createDatabaseWithRights(dbType, dbName, nuxeoUser, nuxeoPW, readerUser, readerPW);
+						createDatabaseWithRights(dbType, dbName, nuxeoUser, nuxeoPW, readerUser);
+						createUtilsSchemaWithRights(dbType, nuxeoUser, repositoryName, cspaceInstanceId);
```
mikejritter, line 1015: The call to create the utils schema. Could also be removed unless it really is necessary.

services/common/src/main/resources/db/postgresql/create_unaccent_text_search_configuration.sql
remillet: The 'create_unaccent_text_search_configuration.sql' script allows text searching to ignore diacritical marks.  This is something that all UCB tenants need.  It's also probably something that ALL CollectionSpace tenants need and should be part of the default CollectionSpace offering.

services/common/src/main/cspace/config/services/tenants/tenant-bindings-proto-unified.xml
ray-lee: Yes, let's remove it.

services/common/src/main/resources/db/postgresql/create_unaccent_text_search_configuration.sql
ray-lee: @remillet But there's nothing that references this script. Is there something missing in this PR?

services/common/src/main/java/org/collectionspace/services/common/ServiceMain.java
ray-lee: I think this is fine to leave in. We can probably find a use for it.

ray-lee (changes requested):

services/common/src/main/java/org/collectionspace/services/common/ServiceMain.java
```diff
@@ -1001,66 +1039,57 @@ private HashSet<String> createNuxeoDatabases() throws Exception {
...
 	 */
 	private void createDatabaseWithRights(DatabaseProductType dbType, String dbName, String ownerName,
-			String ownerPW, String readerName, String readerPW) throws Exception {
-		Connection conn = null;
-		Statement stmt = null;
-		try {
-			DataSource csadminDataSource = JDBCTools.getDataSource(JDBCTools.CSADMIN_DATASOURCE_NAME);
-			conn = csadminDataSource.getConnection();
-			stmt = conn.createStatement();
+			String ownerPW, String readerName) throws Exception {
```
remillet, line 1042: Seems that "ownerPW" is not used.

services/common/src/main/resources/db/postgresql/create_unaccent_text_search_configuration.sql
remillet: I think this script was probably run manually by UCB staff many, many years ago.  It should be part of LYRASIS' tenant-bindings-proto-unified.xml file in the "<tenant:serviceBindings id="runsqlscripts"..." element so that all community instances get this postgresql extension.

services/common/src/main/resources/db/postgresql/botgarden.README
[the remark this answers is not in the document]
remillet: These are UCBG specific, so they don't need to be in the LYRASIS branch.  However, if we want the UC instances to be built off of the LYRAIS code branches, we need a way to provide these kind of UC-specific artifacts.  This is a bigger issue that we should resolve.

remillet (approved):
I've reviewed all changes.  See my comments.

ray-lee (approved):

PR #372, opened 2023-10-18
DRYD-1250: UCB Blob contributions

mikejritter:
**What does this do?**
Brings in changes to the blob api from the UCB contributions:
* Add nuxeo-binary-metadata dependency
* Use binary metadata service for reading blob metadata
* Update logic for checking if a blob is an image
* Add partial content request support to /content

**Additional Changes**
Small changes I made compared to the UCB branch
* Return metadata from the binary metadata service
* Minor formatting updates

Jira: https://collectionspace.atlassian.net/browse/DRYD-1250

This is part of the UCB contributions. 

* Integration Tests:
  * Run the integration test suite with `mvn test`
* Other:
  * Create a media handling record
  * Make a request on /content to test the still api works as expected

This brings in a new dependency, nuxeo-binary-metadata, which replaces the use of the ImagingService in NuxeoBlobUtils#getMetadata. However, the metadata map which is created is never actually used so I'm not sure exactly what the intention is with this (the `imageInfo` is still used, which once again comes from the ImagingService).

Further testing should be done for the partial content to ensure it works correctly. I've only looked at the metadata piece so far.

No.

@mikejritter ran the integration tests and tested the binary metadata service; still needs to check the partial content


ray-lee:
@mikejritter Could you point out where the partial content support is? I don't think I'm seeing it.

services/blob/service/src/main/java/org/collectionspace/services/blob/BlobResource.java
```diff
@@ -256,40 +293,40 @@ public Response create(
...
+    	ServiceContext<PoxPayloadIn, PoxPayloadOut> ctx = null;
+
     	try {
 	    	ctx = createServiceContext(jaxRsRequest, uriInfo);
 			BlobsCommon blobsCommon = getBlobsCommon(csid);
 	    	StringBuffer mimeType = new StringBuffer();
-	    	InputStream contentStream = getBlobContent(ctx, csid, null /*derivative term*/, mimeType /*will get set*/);
-		    
-	    	Response.ResponseBuilder responseBuilder = Response.ok(contentStream, mimeType.toString());
+			File contentFile = getBlobContentFile(ctx, csid, null /*derivative term*/, mimeType /*will get set*/);
```
ray-lee, line 327: What's the motivation for this change?

ray-lee (commented):
This preliminarily looks fine, but I'm not clear on why /content was changed to return a File instead of a Stream. Is it to support partial content requests? If so I'm not seeing where that's implemented.

services/blob/service/src/main/java/org/collectionspace/services/blob/BlobResource.java
remillet: This was to allow partial content requests --which are handled by the RESTEasy/JAX-RS framework when given a File instance.  Was not working when given an InputStream instance.

remillet (approved):
This was to allow partial content requests --which are handled by the RESTEasy/JAX-RS framework when given a File instance. Was not working when given an InputStream instance.

The other change was to remove hard-code MIME types and to let Nuxeo decide whether or not a blob was an image.

services/blob/service/src/main/java/org/collectionspace/services/blob/BlobResource.java
```diff
@@ -149,18 +150,54 @@ public InputStream getBlobContent(ServiceContext<PoxPayloadIn, PoxPayloadOut> ct
...
+
+            PoxPayloadOut response = this.get(csid, ctx);
+            logger.debug(response.toString());
+
+            // The result of a successful get should have put the results in the blobInput instance
+            String mimeType = blobInput.getMimeType();
+            if (mimeType != null) {
+                outMimeType.append(mimeType); // blobInput's mime type was set on call to "get" above by the doc handler
+            }
+            result = BlobUtil.getBlobInput(ctx).getBlobFile();
```
mikejritter, line 184: @ray-lee After doing some testing I found some issues with the blob api from these changes. My first try at testing this was with a plain text file, and when trying to retrieve it I was returned an HTTP 500. I can still see that the Blob has a content stream and I can see the file on disk, so I'm not sure what's going on that the file isn't being set properly. Testing against dev I can retrieve a plain text blob.

Other types of media do work as expected (e.g. jpegs), but we'll want to get this fixed before merging.

ray-lee: Nice find. Thanks for the thorough testing.

mikejritter: I made an update to change `File getBlobContentFile` to `ResponseBuilder getBlobContentResponse`. If a Blob doesn't have a File it will fall back to the InputStream, which won't be able to have the Range header applied and will return the full content stream.

I can make a ticket to track adding partial content support to plain/text. When looking into this yesterday I didn't see a quick way as we're relying on Nuxeo's Blob which doesn't provide a File for certain types.

remillet: @mikejritter @ray-lee When CollectionSpace hands Nuxeo a text/plain file, Nuxeo creates a "Note" document which is backed by Nuxeo's StringBlob class.  The StringBlob class always returns null as the value to the getFileBlob() method.  

If we need/want to support partial content requests for text/plain files, we'll need to change how CollectionSpace asks Nuxeo to create the corresponding blob.

From UCB perspective, there's **no** need to support partial content requests for text/plain files, so Mike's fix should be fine.

ray-lee (approved):
Looks good.

remillet:
@ray-lee @mikejritter 
If we wanted to add support for partial content requests on text/plain files (or other MIME types that Nuxeo doesn't automatically back with a file), we could use this Nuxeo code to get a temp file:

					CloseableFile testClosableFile = docBlob.getCloseableFile();
					File testFile = testClosableFile.getFile();

I did a quick test on the StringBlob and it seems to work.  It creates a temp file here: /usr/local/share/apache-tomcat-8.5.51/nuxeo-server/tmp/

Again, I don't see the need to support partial content on text/plain files, but if you decide the community wants it then this would be a solution.

PR #374, opened 2023-11-07
CB-20: Use controlled object name in computed title

mikejritter:
**What does this do?**
* Add objectNameControlled to denorm'd title

Jira: https://collectionspace.atlassian.net/browse/CB-20

The controlled version of the object name was added in 7.2, but the elasticsearch index wasn't updated to include it in the title. This updates it so that the controlled object name is second in precedence behind the title.

* Deploy the changes and run collectionspace
* Create three collection objects (published to the publicbrowser):
  * One with a title, controlled object name, and object name
  * One with a controlled object name and object name
  * One with an object name
* Query the elasticsearch index and check that the `collectionspace_denorm:title` is correct for each object

Not sure how the index gets updated but the titles will likely be inconsistent until then

No

@mikejritter tested locally


ray-lee (approved):
Looks good.

PR #375, opened 2023-11-07
CB-21: Include controlled terms for material display

mikejritter:
**What does this do?**
* Adds a denormalized materialGroupList to capture controlled materials
* Remove old materialGroupList index

Jira: https://collectionspace.atlassian.net/browse/CB-21

This updates the elasticsearch index to capture the controlled version of a material if present. The field was added in 7.2, so this allows the public browser to be up to date with the latest fields.

- Deploy the changes and run collectionspace
- Create a few collection objects (published to the publicbrowser):
  - One with a controlled material and material
  - One with an uncontrolled material
- Query the elasticsearch index and check that the collectionspace_denorm:materilGropuList is correct for each object, e.g.
```
[elastic] $ cat denorm_query.json
{
  "query": {
    "term": {
      "ecm:primaryType": "CollectionObject"
    }
  },
  "from": 0,
  "size": 23,
  "_source": {
    "includes": [
      "ecm:name",
      "ecm:primaryType",
      "collectionspace_denorm:materialGroupList",
      "collectionspace_denorm:title"
    ]
  },
  "sort": {
    "collectionspace_core:createdAt": {
      "order": "desc"
    }
  },
  "aggs": {
    "material": {
      "terms": {
        "field": "collectionspace_denorm:materialGroupList.material",
        "order": {
          "_term": "asc"
        },
        "size": 300
      }
    }
  }
}
[elastic] $ curl --json @denorm_query.json http://localhost:9200/nuxeo_default/doc/_msearch | jq .
```

Similar to the other public browser PR, I'm not sure how re-indexing will work. With this changing the index it is something that will need to be done so that all materials will display correctly in the filter as well as the display for an object.

I also wasn't sure if we should remove the old `materialGroupList` index or leave it. I've removed it for the tenants which will use the `collectionspace_denorm:materialGroupList` and left it in the material profile. I'm realizing now I should test against the material tenant as well to make sure everything is still working as expected.

No

@mikejritter tested the core tenant locally

### DIFF
--- a/3rdparty/nuxeo/nuxeo-platform-elasticsearch/src/main/java/org/collectionspace/services/nuxeo/elasticsearch/DefaultESDocumentWriter.java
+++ b/3rdparty/nuxeo/nuxeo-platform-elasticsearch/src/main/java/org/collectionspace/services/nuxeo/elasticsearch/DefaultESDocumentWriter.java
@@ -60,6 +60,7 @@ public class DefaultESDocumentWriter extends JsonESDocumentWriter {
 			denormMediaRecords(session, csid, tenantId, denormValues);
 			denormAcquisitionRecords(session, csid, tenantId, denormValues);
 			denormExhibitionRecords(session, csid, tenantId, denormValues);
+			denormMaterialRecords(doc, denormValues);
 
 			// Compute the title of the record for the public browser, and store it so that it can
 			// be used for sorting ES query results.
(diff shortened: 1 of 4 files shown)


3rdparty/nuxeo/nuxeo-platform-elasticsearch/src/main/java/org/collectionspace/services/nuxeo/elasticsearch/DefaultESDocumentWriter.java
```diff
@@ -190,6 +191,37 @@ private void denormExhibitionRecords(CoreSession session, String csid, String te
 	denormValues.putArray("exhibition").addAll(exhibitions);
 }
 
+	/**
+	 * Denormalize the material group list for a collectionobject in order to index the controlled or uncontrolled term
+	 *
+	 * @param doc the collectionobject document
+	 * @param denormValues the json node for denormalized fields
+	 */
+	private void denormMaterialRecords(DocumentModel doc, ObjectNode denormValues) {
```
ray-lee, line 200: I think this should be called `denormMaterialFields`, as it's just combining fields into one, rather than pulling fields from related records.

services/common/src/main/cspace/config/services/tenants/tenant-bindings-proto-unified.xml
```diff
@@ -1230,6 +1229,16 @@
 								}
 							}
 						},
+						"collectionspace_denorm:materialGroupList": {
+							"type": "object",
+							"properties": {
+								"material": {
+									"type": "keyword",
+									"copy_to": "all_field",
+									"normalizer": "sorting_normalizer"
```
ray-lee, line 1238: Is there a reason for using `sorting_normalizer`? I hadn't been using it for facets.

ray-lee (commented):
Oh, I think this is actually CB-19. I'm not sure, but it might be saying to index both values, rather than one or the other. You might want to check with Jessi.

kspurgin:
With both the objectname and material updates for the public browser, we had discussed that both the freetext field value (e.g. material) and controlled field value (e.g. materialControlled) should be extracted and indexed into the same facet field in the public browser. 

*Usually* people are going to either be populating one or the other (freetext or controlled) in their instance, but if somebody did, in the same record, put materialControlled: "gold" and material: "24 carat gold", then both "gold" and "24 carat gold" should be terms in the Material facet, and both terms' results should include the record that contains the two terms in the two fields.

mikejritter:
> Oh, I think this is actually (or, also) CB-19. I'm not sure, but it might be saying to index both values, rather than one or the other. You might want to check with Jessi.

This work fits into CB-19 better - I was initially thinking of doing CB-19 + CB-20 in one PR (which would be CB-21) but separated them out.

I just pushed a few commits to cover the comments made. Notably to capture both the uncontrolled and controlled material fields as well as remove the sorting normalizer.

ray-lee (approved):
Looks good!

PR #378, opened 2023-11-29
DRYD-1328: Use distinct in order to get media for all rows

mikejritter:
**What does this do?**
* Use distinct in order to get media for all csids

Jira: https://collectionspace.atlassian.net/browse/DRYD-1328

This was a bug identified when selecting multiple collection objects from the tombstone with budget and tombstone with creator reports. It is caused by the subquery returning a single row instead of one row per object because of using `LIMIT`. To fix this I updated the query to use `DISTINCT` on each csid.

* Rebuild collectionspace with the publicart tenant enabled
* Start collectionspace and run the publicart ui
* Create multiple collectionobjects with associated media
  * Using a selection of them, run the `Tombstone with Budget` report with pdf output
  * See that each collectionobject has a thumbnail
  * Using a selection of them, run the `Tombstone with Creator` report with pdf output
  * See that each collectionobject has a thumbnail

None

No

@mikejritter tested locally


ray-lee (approved):
Looks good.

PR #380, opened 2023-12-06
DRYD-1249: UCB Structured Date Contrib

mikejritter:
**What does this do?**
* Updates to exitRomanStringDate
* Add additional test dates
* Remove duplicate condition

Jira: https://collectionspace.atlassian.net/browse/DRYD-1249

This is part of the UCB contributions and brings in changes to structured date from their branch. It makes changes to `romanDate` which is now mapped to `romanStringDate`. There was also a duplicate condition which I removed.

I'm not too familiar with structured dates and what's expected of them so I only did some basic testing:
* Build collectionspace
* Run integration tests
* Create a structured date
  * This should probably be done with a roman formatted date as well

?

No

### DIFF
--- a/services/structureddate/structureddate/src/main/antlr4/org/collectionspace/services/structureddate/antlr/StructuredDate.g4
+++ b/services/structureddate/structureddate/src/main/antlr4/org/collectionspace/services/structureddate/antlr/StructuredDate.g4
@@ -63,7 +63,7 @@ date:                  numDate
 |                      dayFirstDate
 |                      dayOrYearFirstDate
 |                      invStrDateEraLastDate
-|                      romanDate
+|                      romanStringDate
 ;
 
 month:                 monthYear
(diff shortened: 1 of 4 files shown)


kspurgin:
It also seems odd/surprising to be able to handle `1908-IX-01` and `1908-01-IX`, but not `1908-IX-I`or `MCMVIII-IX-I`. As a user, it seems odd to me that only months can be expressed in Roman numerals. 

(Of course, expressing the individual segments of a YYYY-MM-DD or YYYY-DD-MM date in Roman numerals at all seems extremely odd, but if we are going to be odd, we should be odd consistently)

kspurgin:
@abbistani Figured it must be some obscure use case like that. Mentally filed. 

BUT I stand by thinking it is weird/surprising system behavior to only support Roman numerals in month segments, but not as a standalone year value

ray-lee (approved):
This looks fine. I'll make a ticket for adding support for roman years.

PR #381, opened 2023-12-06
DRYD-1251: UCB ImageMagick Config Contrib

mikejritter:
**What does this do?**
* Config for setting ImageMagick path

Jira: https://collectionspace.atlassian.net/browse/DRYD-1251

This adds additional build configuration so that the ImageMagick path can be set in order to control which install is used. 

* Build collectionspace with `cspace.im.root` set to a local install
* Run collectionspace and see that the correct version of imagemagick is used

Additional bundle requirement should likely be double checked.

No, imagemagick docs and anything going over build properties will need to be updated to reflect new config option.


ray-lee:
@remillet Why do you need this?

ray-lee (changes requested):
This looks fine, but I'm missing the use case/context for why it's needed (@remillet). Also @remillet please provide documentation on how to maintain this when Nuxeo is upgraded, including where the contents of commandline-imagemagick-contrib.xml come from.

remillet:
@ray-lee UCB needed to add this because the bIT (Berkeley IT) supplied base OS was using an incompatible version of ImageMagick.  We (RTL) had to install our own version of ImageMagick and configure Nuxeo to use that version.

LYRASIS hosting may not have that problem, but others in the community may run into this issue.

The documentation from Nuxeo is here: [CommandLineExecutorComponent](https://explorer.nuxeo.com/nuxeo/site/distribution/Nuxeo%20Platform%20LTS%202017-9.10/viewComponent/org.nuxeo.ecm.platform.commandline.executor.service.CommandLineExecutorComponent)

remillet (approved):
See my comments.

ray-lee:
@remillet I still don't understand where the contents of commandline-imagemagick-contrib.xml came from. How did you know what the command names should be, and the parameters, etc.? I'm not seeing that in the documentation you linked.

ray-lee:
Aha, I found the correct documentation: https://explorer.nuxeo.com/nuxeo/site/distribution/Nuxeo%20Platform%20LTS%202017-9.10/viewContribution/org.nuxeo.ecm.platform.picture.commandline.imagemagick--command

ray-lee (approved):
Added the documentation to a comment in commandline-imagemagick-contrib.xml.

PR #383, opened 2023-12-07
DRYD-1268: UCB Maven Contribs

mikejritter:
**What does this do?**
* Add build-helper-maven-plugin
* Add flatten-maven-plugin to pluginConfiguration
* Update tomcat version number
* Dependency and plugin updates
* Add jaxb-api to hyperjaxb, config, common, authorization jaxb, and authorization-mgt services modules
* Javassist exclusion for services common (jaxrs service provider)
* Xalan exclusion for module-xml_merge-common (root)
* Remove extraneous .class files
* Remove commented out dependencies/blocks from some poms
* Remove unused 'sample' profile
* Remove empty plugin config blocks

Jira: https://collectionspace.atlassian.net/browse/DRYD-1268

These are maven contributions from the UCB bundle. It includes various housekeeping done to various poms and source files, as well as some dependency/plugin updates.

##### Plugins
* flatten-maven-plugin was added to the pluginConfiguration section, so no real impact on the build
* build-helper-maven-plugin which adds the `generated-sources/xjc` directory to each project. One thing I noted is that the source files in `generated-sources/xjc` are already being added to each jaxb artifact, so it's not clear if this is strictly necessary. 

##### Dependency/Plugin Updates
Not much here, just a few minor version increments for maven-antrun-plugin and slf4f-api. IIRC there were some others which had fallen behind our recent dependency updates.

The biggest change here is adding the jaxb-api dependency to various modules (outlined above). From what I saw there are imports from jaxb-api so this makes sense to have. I only did a quick check, and I could look again to verify if needed.

##### Housekeeping
Some minor changes were made to remove commented out blocks from various pom files. I also took this as an opportunity to remove some unused profiles (notably, sample) and empty plugin configuration.

There were also some .class files which were removed from the source tree, which I bundled in with these changes.

The tomcat version bump also seems to fit with housekeeping, though we might want to update it again if we've updated the version we deploy against. 

* Build and deploy collectionspace
* Run integration tests

Since these are mostly changes to the build and nothing functional, if collectionspace starts and tests pass that should indicate everything is ok.

A few things to look over before merging, as outlined in the 'why we're doing this' section

No


ray-lee (approved):
This looks ok to me.

PR #384, opened 2023-12-07
DRYD-1268: UCB Misc/QoL Contribs

mikejritter:
**What does this do?**
* Add failCtxIfServletStartFails to context.xml
* Add null check to CSpaceUserDetails#loadUserByUsername
* Add null check to SpringAuthNContext#getCurrentTenantId
* Add null check to ServiceBindingUtils#getAllPartsPropertyValues
* Add empty check to AbstractServiceContextImpl#createDocumentHandlerInstance
* Add missing Override annotation for RelationResource#getList
* Add new parameters for JpaStorageUtils#getAccountRoles
* Lowercase relation affects/broader constants
* Change password used in EmailUtil
* Log sql when create new db user fails
* Use try-with-resource when creating new db user
* Add stopNuxeoEP to attempt deregistration of drivers when shutting down
* Log error when failing to start NuxeoFramework
* Whitespace formatting for PropertyItemUtils#getPropertValuesByName
* Remove duplicate license in CSpaceUserDetails

Jira: https://collectionspace.atlassian.net/browse/DRYD-1268

This is the second part of the misc UCB contributions. These are mostly small quality of life changes, revolving around some null checks, logging updates, and other minor changes. 

I made some additional changes on top, but generally only where changes were already being made. This includes additional formatting, javadoc cleanup, and using try w/ resources where possible. I committed everything individually in case there's anything we end up wanting to roll back.

Although these changes are small, they cover a wide variety of things which we should test, starting with the integration tests:
* Build and run collectionspace
* Run integration tests
* Test access of collectionspace with logged in user (SpringAuthNContext)
  * If possible it would be nice to test an account w/o any tenant association, which I think we might need to manually alter the database to do (CSpaceUserDetailsService)
* Create/Delete a relation on an object/procedure (RelationResource/RelationConstants)
* Create a new user (JDBCTools)
* View roles for a user (JpaStorageUtils)
* Shutdown collectionspace (NuxeoConnectorEmbedded)

A few things which I want to review before merging:
* RelationConstants lowercase change
  * The constants are only used in a few places, so it might be good to do some refactoring there. I noticed for broader it's doing an equals, so we might want to ignore case. Or we can switch to check against the enum types, which would be mostly the same.   
* JpaStorageUtils getAccountRoles additional parameters
  * The only usage of this is the call passing in null arguments. Should probably be removed.
* EmailUtil
  * IMO it would be good to review this in general and scrub it of usernames/passwords
  * Also if any sensitive information was in here previously, it would probably be a good time to scrub it from the repo (maybe instead of force pushing master, using this as an opportunity to push to main without the commits)
* NuxeoFrameworkBootstrap 
  * Wondering if we can switch from System.warn to using a logger. I'll try this on my own after creating the PR.

No

### DIFF
--- a/services/common/src/main/java/org/collectionspace/services/common/relation/nuxeo/RelationConstants.java
+++ b/services/common/src/main/java/org/collectionspace/services/common/relation/nuxeo/RelationConstants.java
@@ -54,6 +54,6 @@ public class RelationConstants {
     public final static String TYPE_SCHEMA_NAME = COMMON_SCHEMA_NAME;
     public final static String TYPE_FIELD_NAME = IRelationsManager.RELATIONSHIP_TYPE;
     
-    public final static String AFFECTS_TYPE = RelationshipType.AFFECTS.toString();
-    public final static String BROADER_TYPE = RelationshipType.HAS_BROADER.toString();
+    public final static String AFFECTS_TYPE = RelationshipType.AFFECTS.toString().toLowerCase();
+    public final static String BROADER_TYPE = RelationshipType.HAS_BROADER.toString().toLowerCase();
 }
--- a/services/common/src/main/java/org/collectionspace/services/common/relation/nuxeo/RelationConstants.java
+++ b/services/common/src/main/java/org/collectionspace/services/common/relation/nuxeo/RelationConstants.java
@@ -54,6 +54,6 @@ public class RelationConstants {
     public final static String TYPE_SCHEMA_NAME = COMMON_SCHEMA_NAME;
     public final static String TYPE_FIELD_NAME = IRelationsManager.RELATIONSHIP_TYPE;
     
-    public final static String AFFECTS_TYPE = RelationshipType.AFFECTS.toString().toLowerCase();
-    public final static String BROADER_TYPE = RelationshipType.HAS_BROADER.toString().toLowerCase();
+    public final static String AFFECTS_TYPE = RelationshipType.AFFECTS.value();
+    public final static String BROADER_TYPE = RelationshipType.HAS_BROADER.value();
 }
--- a/services/common/src/main/java/org/collectionspace/services/common/storage/JDBCTools.java
+++ b/services/common/src/main/java/org/collectionspace/services/common/storage/JDBCTools.java
@@ -606,6 +606,8 @@ public class JDBCTools {
         if (dbType != DatabaseProductType.POSTGRESQL) {
             throw new UnsupportedOperationException("createNewDatabaseUser only supports PostgreSQL");
         }
+        
+        String sql = null;
         try {
             if (hasDatabaseUser(dataSourceName, repositoryName, cspaceInstanceId, dbType, username)) {
                 if (logger.isDebugEnabled()) {
@@ -614,7 +616,7 @@ public class JDBCTools {
             } else {
                 conn = getConnection(dataSourceName, repositoryName, cspaceInstanceId);
                 stmt = conn.createStatement();
-                String sql = "CREATE ROLE " + username + " WITH PASSWORD '" + userPW + "' LOGIN";
+                sql = "CREATE ROLE " + username + " WITH PASSWORD '" + userPW + "' LOGIN";
                 stmt.executeUpdate(sql);
                 if (logger.isDebugEnabled()) {
                     logger.debug("Created User: " + username);
@@ -622,6 +624,9 @@ public class JDBCTools {
             }
         } catch (Exception e) {
             logger.error("createNewDatabaseUser failed on exception: " + e.getLocalizedMessage());
+            if (sql != null) {
+            	logger.error(String.format("The following SQL statement failed using credentials from datasource named '%s': '%s'", dataSourceName, sql));
+            }
             throw e;
         } finally {
             try {
--- a/services/common/src/main/java/org/collectionspace/services/common/storage/JDBCTools.java
+++ b/services/common/src/main/java/org/collectionspace/services/common/storage/JDBCTools.java
@@ -617,10 +617,6 @@ public class JDBCTools {
             }
         } catch (Exception e) {
             logger.error("createNewDatabaseUser failed with exception:", e);
-            if (sql != null) {
-                logger.error("The following SQL statement failed using credentials from datasource named '{}': '{}'",
-                             dataSourceName, sql);
-            }
             throw e;
         }
     }
--- a/services/common/src/main/java/org/collectionspace/services/common/storage/JDBCTools.java
+++ b/services/common/src/main/java/org/collectionspace/services/common/storage/JDBCTools.java
@@ -625,7 +625,8 @@ public class JDBCTools {
         } catch (Exception e) {
             logger.error("createNewDatabaseUser failed on exception: " + e.getLocalizedMessage());
             if (sql != null) {
-                logger.error(String.format("The following SQL statement failed using credentials from datasource named '%s': '%s'", dataSourceName, sql));
+                logger.error("The following SQL statement failed using credentials from datasource named '{}': '{}'",
+                             dataSourceName, sql);
             }
             throw e;
         } finally {
--- a/services/common/src/main/java/org/collectionspace/services/common/storage/JDBCTools.java
+++ b/services/common/src/main/java/org/collectionspace/services/common/storage/JDBCTools.java
@@ -594,7 +594,9 @@ public class JDBCTools {
     /**
      * Create a database user, if that user doesn't already exist.
      *
-     * @param conn a database connection.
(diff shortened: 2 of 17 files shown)


services/common/src/main/java/org/collectionspace/services/common/relation/nuxeo/RelationConstants.java
```diff
@@ -54,6 +54,6 @@ public class RelationConstants {
     public final static String TYPE_SCHEMA_NAME = COMMON_SCHEMA_NAME;
     public final static String TYPE_FIELD_NAME = IRelationsManager.RELATIONSHIP_TYPE;
     
-    public final static String AFFECTS_TYPE = RelationshipType.AFFECTS.toString();
-    public final static String BROADER_TYPE = RelationshipType.HAS_BROADER.toString();
+    public final static String AFFECTS_TYPE = RelationshipType.AFFECTS.toString().toLowerCase();
+    public final static String BROADER_TYPE = RelationshipType.HAS_BROADER.toString().toLowerCase();
```
ray-lee, line 58: These smell funny. Are we sure about this? It would make sense to me if they were `RelationshipType.AFFECTS.value()` and `RelationshipType.HAS_BROADER.value()`, but `toLowerCase()` seems arbitrary. It looks like these constants are only used by a botgarden listener and a botgarden batch job, therefore only by UCB, so if things are working as you expect, @remillet, I guess its fine? But are you really sure you want this?

services/common/src/main/java/org/collectionspace/services/common/storage/JDBCTools.java
```diff
@@ -601,40 +601,27 @@ public static void grantPrivilegeToDatabaseUser(String dataSourceName, String re
...
-                if (conn != null) {
-                    conn.close();
-                }
-            } catch (SQLException sqle) {
-                // nothing we can do here except log
-                logger.warn("SQL Exception when closing statement/connection: " + sqle.getLocalizedMessage());
+            logger.error("createNewDatabaseUser failed with exception:", e);
+            if (sql != null) {
+                logger.error("The following SQL statement failed using credentials from datasource named '{}': '{}'",
+                             dataSourceName, sql);
```
ray-lee, line 622: I'm not comfortable logging a user-submitted password. I think I'd rather not know the SQL statement than possibly accidentally see someone's intended password. Even if the user failed to be created now, the password might be reused when it succeeds.

services/common/src/main/java/org/collectionspace/services/common/storage/jpa/JpaStorageUtils.java
```diff
@@ -335,6 +335,12 @@ public static AccountPermission getAccountPermissions(String csid, String curren
 
 	//FIXME: REM - This method should probably be moved to the AccountPermissionDocumentHandler
 	public static AccountRole getAccountRoles(String csid)
+		throws UnauthorizedException, DocumentNotFoundException {
+		return getAccountRoles(csid, null, null);
+	}
+
+	//FIXME: REM - This method should probably be moved to the AccountPermissionDocumentHandler
+	public static AccountRole getAccountRoles(String csid, String currentResource, String permissionResource)
```
ray-lee, line 343: `String currentResource, String permissionResource` were added to this method signature, but they are never used. Any idea why?

services/common/src/main/java/org/collectionspace/services/common/storage/jpa/JpaStorageUtils.java
```diff
@@ -375,18 +381,12 @@ public static AccountRole getAccountRoles(String csid)
...
-				AccountRoleRel accountRolRel = (AccountRoleRel) resultList.next();
-				RoleValue roleValue = AuthorizationRoleRel.buildRoleValue(accountRolRel);
-
-				if (roleValue != null) {
-					roleValues.add(roleValue);
+			if (resultList.hasNext()) {
+				List<RoleValue> roleValues = new ArrayList<RoleValue>();
+				while (resultList.hasNext()) {
+					AccountRoleRel accountRolRel = (AccountRoleRel)resultList.next();
+					roleValues.add(AuthorizationRoleRel.buildRoleValue(accountRolRel));
```
ray-lee, line 388: We lost a null check on `AuthorizationRoleRel.buildRoleValue(accountRolRel)`. Is that intentional?

ray-lee (changes requested):

services/common/src/main/java/org/collectionspace/services/common/storage/JDBCTools.java
mikejritter: When looking at this initially I'm pretty sure I mixed up how this query is made w/ a PreparedStatement. Given that this is an error level too it's almost guaranteed to show up in the logs. I'll remove it from the PR.

Honestly I'd like to change how the query is built too, I think just concatenating the Strings makes it a candidate for injection.

ray-lee: That is an excellent point. It would be great to change to a PreparedStatement.

mikejritter: @ray-lee While making changes I realized that this is a DDL statement and can't be used with a PreparedStatement. It's only called when creating the nuxeo database, and it seems to be getting the user/password from the DataSource created by the webapp context, so it's not _really_ user input. 

Likely something to think more about though, especially in terms of how we want to handle db initialization in the future.

ray-lee: Ah, ok. Let's still not log the password, in case someone who has access to the logs shouldn't have access to the datasource configuration.

services/common/src/main/java/org/collectionspace/services/common/storage/jpa/JpaStorageUtils.java
remillet: @ray-lee I have no idea why this change happened.  If I'm interpreting git's 'blame' correctly, it may be a change you made several years ago?  See https://github.com/cspace-deployment/services/commit/28390c0771ad1bbe743ed743d98f0d344c846845

Either way, it doesn't seem necessary so we should stick with the current LYRASIS 'master' branch version.

services/common/src/main/java/org/collectionspace/services/common/storage/jpa/JpaStorageUtils.java
remillet: Not intentional.  Looks like old code from before a change @ray-lee made several years ago.

services/common/src/main/java/org/collectionspace/services/common/relation/nuxeo/RelationConstants.java
remillet: Sure, RelationshipType.AFFECTS.value() should work.

Looks like toString() was returning "AFFECTS" and value() returns "affects".  Lots of places in the code (including parts of the UI?) assume lowercase "affects".

services/common/src/main/java/org/collectionspace/services/nuxeo/client/java/NuxeoConnectorEmbedded.java
```diff
@@ -147,6 +154,31 @@ private void startNuxeoEP(String serverRootPath) throws Exception {
...
+			while (drivers.hasMoreElements()) {
+				Driver driver = drivers.nextElement();
+				try {
+					DriverManager.deregisterDriver(driver);
+					logger.info("Deregister JDBC driver: {}", driver);
+				} catch (SQLException e) {
+					logger.error("Error deregistering JDBC driver {}", driver, e);
+				}
+			}
+		} catch (IllegalArgumentException | ReflectiveOperationException | JMException e) {
```
remillet, line 172: Why not log the exception from this catch block?

services/common/src/main/java/org/collectionspace/services/nuxeo/client/java/NuxeoFrameworkBootstrap.java
```diff
@@ -47,7 +48,11 @@ public void start(MutableClassLoader cl) throws ReflectiveOperationException, IO
             throw new IllegalStateException("Framework Loader was not initialized. Call initialize() method first");
         }
         Method start = frameworkLoaderClass.getMethod("start");
-        start.invoke(null);
+        try {
+            start.invoke(null);
+        } catch (BlobNotFoundException e) {
```
remillet, line 53: Huh? This does seem to make sense.  Don't include this change.  BlobNotFoundException will never happen.

remillet (approved):
I've reviewed this and have left comments.

ray-lee (approved):
Looks good.

PR #385, opened 2024-01-11
DRYD-1370: Join on most recent media

mikejritter:
**What does this do?**
This updates the joins for `related_media` queries in reports to restrict to a single row of the most recently updated media.

Jira: https://collectionspace.atlassian.net/jira/software/c/projects/DRYD/issues/DRYD-1370

It was determined that returning all related media was not desirable, and that a single thumbnail/csid was preferred. 

* Rebuild collectionspace
* Create 2 Media Handling for testing
* Create a CollectionObject with relations to both Media Handling procedures
* **Basic Object with Current Location**
  * Search on CollectionObjects and run `Basic Object with Current Location` report
  * Verify the thumbnails are correct for each CollectionObject in the report
* **Accessions Test**
  * Create an Acquisition for the CollectionObject
  * Run the `Accessions` report for the Acquisition and verify the thumbnail is the most recent media
  * _Note: Since Acquisitions can be related to multiple CollectionObjects, trying 2 or 3 could be desirable_
* **Deaccessions Test**
  * Create an ObjectExit for the CollectionObject
  * Run the `Deaccessions` report for the ObjectExit and verify the thumbnail is the most recent media
* **Condition Test**
  * Create a ConditionCheck for the CollectionObject
  * Related one or both of the Media Handling records to the Condition Check
  * Search for CollectionObjects and run the `Condition` report
  * Verify the thumbnails are correct for each CollectionObject

None

No

@mikejritter tested the reports on a local install

### DIFF
--- a/services/report/3rdparty/jasper-cs-report/src/main/resources/accessions.jrxml
+++ b/services/report/3rdparty/jasper-cs-report/src/main/resources/accessions.jrxml
@@ -83,12 +83,15 @@
   LEFT JOIN hierarchy opdg_hier ON opdg_hier.parentid = object.id AND opdg_hier.name = 'collectionobjects_common:objectProductionDateGroupList' AND opdg_hier.pos = 0
   LEFT JOIN structureddategroup sdg ON sdg.id = opdg_hier.id
 ), related_media AS (
-  SELECT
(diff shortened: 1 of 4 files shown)


ray-lee (approved):
Looks good.

PR #388, opened 2024-01-29
DRYD-1373: Update reports that use numberofobjects

mikejritter:
**What does this do?**
This removes the usage of `numberofobjects` in reports and performs a join on the new `objectcountgroup` table in order to use `objectcount`.

Jira: https://collectionspace.atlassian.net/browse/DRYD-1373

As part of replacing the `numberofobjects` field, we need to update references to it in various places. This step is for updating the reports which rely on that field so that they can continue to be used with the correct information.

* Deploy collectionspace
* Create a CollectionObject with one or more `objectCount`s.
* Run the following reports:
  * Acquisition, Acquisition Basic List - requires a related collectionobject
  * Condition Check, ConditionCheck Basic List - requires a related collectionobject
  * Exhibition, Exhibition Basic List - requires a related collectionobject
  * Loan In, Loan In Basic List - requires a related collectionobject
  * Loan Out, Loan Out Basic List - requires a related collectionobject
  * CollectionObject, Group Basic List
* In the publicart profile run the following report
  * CollectionObject, Full Object with Place Details 

I wasn't sure on the context for running the `catalogue_sheet` and `systematicInventory` reports. The xml for `catalogue_sheet` is missing thought looking at the report I could likely just do it on a single collectionobject.

The `systematicInvetory` only has `supportsNoContext` set, which I wasn't sure about.

No

@mikejritter ran the reports with an objectcount set

### DIFF
--- a/services/report/3rdparty/jasper-cs-report/src/main/resources/Acq_List_Basic.jrxml
+++ b/services/report/3rdparty/jasper-cs-report/src/main/resources/Acq_List_Basic.jrxml
@@ -32,7 +32,7 @@
 						ptg.termdisplayname opperson,
 						objectproductionpeople,
 						datedisplaydate,
-						numberofobjects,
+						objectCountGroup.objectcount numberofobjects,
 						dimensionsummary,
 						computedcurrentlocation
 	FROM acquisitions_common ac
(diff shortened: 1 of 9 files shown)


ray-lee (approved):
Looks good.

PR #389, opened 2024-01-31
DRYD-1378: Updates for objectCountGroup

mikejritter:
**What does this do?**
* Add objectCountGroupList to the collectionobject jaxb schema
* Add objectCountGroupList to es included fields
* Add objectCountGroupList to xml templates for testing
* Whitespace cleanup in CollectionObjectServiceTest

Jira: https://collectionspace.atlassian.net/browse/DRYD-1378

This adds objectCountGroupList and related fields where `numberOfObjects` is used in the services layer. It does not remove numberOfObjects but lets them live side by side until the migration is complete.

* Rebuild collectionspace
* Run the integration tests

Some of the xml files do not appear to be used, so some of this is more for completeness than anything else (e.g. `services/imports/.../collectionobject-request.xml`). 

I'm also not 100% certain how numberOfObjects is used in the material public browser. I was going to dig into it more with the public browser ticket.

No

@mikejritter tested against a local server. I have an integration test which is failing that I'm looking into, though appears at a glance to be unrelated.


ray-lee (approved):
Looks good.

PR #394, opened 2024-02-27
CB-26: Add denormalized field for content fields

mikejritter:
**What does this do?**
* Creates aggregate field for indexing content fields

Jira: https://collectionspace.atlassian.net/jira/software/c/projects/CB/issues/CB-26

This work is to help simplify the public browser logic for displaying these related fields under a single heading. It also resolves the issue of indexed items being in the detail display for an object but not being seen in the filter panel.

* Rebuild collectionspace with elasticsearch enabled
* Create a collectionobject with the content concept, organization, event, persons, and description filled out
* Query elasticsearch for the denormalized field, e.g.
```
[publicbrowser] $ cat subject.json 
{"preference":"result"}
{"query":{"term":{"ecm:primaryType":"CollectionObject"}},"from":0,"size":23,"_source":{"includes":["ecm:name","ecm:primaryType","collectionspace_denorm:contentSubjectList"]},"sort":{"collectionspace_core:createdAt":{"order":"desc"}},"aggs":{"subject":{"terms":{"field":"collectionobjects_denorm:contentSubjectList.subject","order":{"_term":"asc"},"size":300}}}}
[publicbrowser] $ curl -H "Content-Type: application/x-ndjson" -XGET http://localhost:9200/nuxeo_default/doc/_msearch --data-binary @subject.json | jq
```

Two things before merging:
* I was considering adding an extra field so that we can determine if the subject is able to be linked to in the public browser. I believe the description is the only field which wouldn't be linked to, but I'm still testing the public browser side of things.
  * The link for the description actually appears to work fine, so this doesn't seem necessary anymore
* I noticed the tenant bindings for materials was missing some of the newer denorm fields, e.g. objectNameList. I _think_ we'll want to add that and the contentSubjectList, but didn't make changes yet.

No

### DIFF
--- a/3rdparty/nuxeo/nuxeo-platform-elasticsearch/src/main/java/org/collectionspace/services/nuxeo/elasticsearch/DefaultESDocumentWriter.java
+++ b/3rdparty/nuxeo/nuxeo-platform-elasticsearch/src/main/java/org/collectionspace/services/nuxeo/elasticsearch/DefaultESDocumentWriter.java
@@ -255,7 +255,7 @@ private void denormExhibitionRecords(CoreSession session, String csid, String te
 	}
 
 	/**
-	 * Denormalize the content concept, content event, content person, content organization, and content description
+	 * Denormalize the content concept, content event, content person, and content organization
 	 * fields for a collectionobject so that they are indexed under a single field
 	 *
 	 * @param doc the collectionobject document
@@ -278,13 +278,6 @@ private void denormExhibitionRecords(CoreSession session, String csid, String te
 					denormContentSubject.add(node);
 				}
 			}
-		}
-
-		String description = (String) doc.getProperty("collectionobjects_common", "contentDescription");
-		if (description != null) {
-			final ObjectNode node = objectMapper.createObjectNode();
-			node.put("subject", description);
-			denormContentSubject.add(node);
 		}
 
 		denormValues.putArray("contentSubjectList").addAll(denormContentSubject);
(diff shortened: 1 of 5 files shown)


3rdparty/nuxeo/nuxeo-platform-elasticsearch/src/main/java/org/collectionspace/services/nuxeo/elasticsearch/DefaultESDocumentWriter.java
```diff
@@ -253,6 +254,42 @@ private void denormObjectNameFields(DocumentModel doc, ObjectNode denormValues)
...
+			for (String content  : contentList) {
+				if (content != null) {
+					final ObjectNode node = objectMapper.createObjectNode();
+					node.put("subject", RefNameUtils.getDisplayName(content));
+					denormContentSubject.add(node);
+				}
+			}
+		}
+
+		String description = (String) doc.getProperty("collectionobjects_common", "contentDescription");
```
mikejritter, line 283: @ray-lee looking back at the discussion about the content fields I'm not sure if we need this. I'm going to message Jessi about the display again to double check. I'll add an update here + in jira once I get a response.

mikejritter:
@ray-lee After talking with Jessi I determined that the content description wasn't required to be displayed under the `Subject` heading with the other content fields. I've removed it from the denorm field and updated this and the public browser PR.

ray-lee (approved):
Looks good.

PR #396, opened 2024-02-29
Deurn acquisitionmethod in accessions report

mikejritter:
**What does this do?**
* Adds acquisitionmethod to deurn fields

Jira: https://collectionspace.atlassian.net/browse/DRYD-1129

When running this report in the publicart profile it was seen that acquisitionmethod was returning the cspace urn. This updates the report so that the acquisitionmethod field is included with the deurnfields so that it uses its display name. 

* Run the acquisitions report on the publicart profile
* See that the acquisitionmethod field is deurn'd

No

None

@mikejritter tested


ray-lee (approved):
Looks good.

PR #397, opened 2024-02-29
Add placenote to full object with place report

mikejritter:
**What does this do?**
This adds the place note to the full object with place report

Jira: https://collectionspace.atlassian.net/browse/DRYD-1281

After testing in QA it was requested to add place note as a field to this report.

* Run the publicart ui devserver
* Create a collectionobject with a related place (from the currentcomputedlocation)
* Run the full object with place report
* Verify that the place note exists for the report

None

No

@mikejritter tested


mikejritter:
Just a quick note -- I'm in the process of testing this and #396. I think I ran them last week but didn't document anything so I want to rerun the reports.

ray-lee (approved):
Looks good.

PR #399, opened 2024-02-29
Update Tombstone with Creator Report

mikejritter:
**What does this do?**
This adds the production date type to the artwork date field

Jira: https://collectionspace.atlassian.net/browse/DRYD-1196

  This change was requested for the public art profile to modify the response for the artwork date field. It brings in the production date type which is the adjacent field in the group.

* Using the publicart profile
  * Artwork date and type (Production Information)
* Search for objects and run the Tombstone with Creator Report

None

No


ray-lee (approved):
Looks good.

PR #402, opened 2024-03-07
Add migration script for collectionobject fieldCollectionPlaces

mikejritter:
**What does this do?**
Create a script for migrating collectionobjects_common.fieldCollectionPlace into collectionobjects_common_fieldcollectionplaces. This allows for the field to be repeatable.

Jira: https://collectionspace.atlassian.net/browse/DRYD-1395

The fieldCollectionPlace field was requested to be updated to be a repeating field. Any data from 7.2 needs to be migrated and inserted it into the new table collectionobjects_common_fieldcollectionplaces. 

* Using collectionspace 7.2, create a collectionobject with a fieldCollectionPlace
  * As a side note - I had to update cspace.meta because I built with 8.0.0-SNAPSHOT 
* Stop collectionspace
* Using this PR, rebuild collectionspace and start the server
* Check the database to see that the data has been migrated
```
nuxeo_default=# select * from collectionobjects_common_fieldcollectionplaces;
```
* Or check through the cspace-ui PR which updates the ui to show the field in a repeatable display

Will need to make sure the other migration script needed for 8.0 does not conflict with this one. It will also be on collectionobject so likely will happen.

No

### DIFF
--- a/src/main/resources/db/postgresql/upgrade/8.0.0/post-init/01_collectionobject.sql
+++ b/src/main/resources/db/postgresql/upgrade/8.0.0/post-init/01_collectionobject.sql
@@ -1,0 +1,33 @@
+-- Upgrade collectionobject. Move fieldCollectionPlace into repeating field (DRYD-1395).
+DO $$
+DECLARE
+    trow record;
+    maxpos int;
+BEGIN
+    -- For new install, if collectionobjects_common.fieldcollectionplace does not exist, there is nothing to migrate.
+    IF EXISTS (SELECT 1 FROM information_schema.columns WHERE table_name='collectionobjects_common' AND column_name='fieldcollectionplace') THEN
+        FOR trow IN
+            -- Get record in collectionobjects_common that does not have an existing/matching record in
+            -- collectionobjects_common_fieldcollectionplaces:
+            SELECT co.id, co.fieldcollectionplace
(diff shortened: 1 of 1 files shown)


src/main/resources/db/postgresql/upgrade/8.0.0/post-init/01_collectionobject.sql
```diff
@@ -0,0 +1,33 @@
...
+            -- Get record in collectionobjects_common that does not have an existing/matching record in
+            -- collectionobjects_common_fieldcollectionplaces:
+            SELECT co.id, co.fieldcollectionplace
+            FROM public.collectionobjects_common co
+            LEFT OUTER JOIN public.collectionobjects_common_fieldcollectionplaces co_fcp ON
+                (co.id = co_fcp.id AND co.fieldcollectionplace = co_fcp.item)
+            WHERE co.fieldcollectionplace IS NOT NULL AND co.fieldcollectionplace != '' AND co_fcp.item IS NULL
+
+            LOOP
+                -- Get max pos value for the collectionobject record's current owner field:
```
ray-lee, line 19: Looks like you missed updating this comment after copy/paste. "current owner" should be "field collection place".

ray-lee (changes requested):

ray-lee (approved):
Looks good.

PR #405, opened 2024-04-30
Add DutyOfCare

mikejritter:
**What does this do?**
* Creates the DutyOfCare procedure

Jira: https://collectionspace.atlassian.net/browse/DRYD-1332

This is one of the procedures for the NAGPRA support we're working on for the next release.

* Rebuild collectionspace
* Run the test suite against collectionspace

Not a dependency but I've been playing around with using spotless to handle formatting. I added it to the parent dutyofcare pom, and it can be run with `mvn spotless:check` or `mvn spotless:apply`. I'm only doing it for the procedures for now as applying to everything is slightly overwhelming.

No

@mikejritter tested the procedure through the ui updates

### DIFF
--- a/services/JaxRsServiceProvider/pom.xml
+++ b/services/JaxRsServiceProvider/pom.xml
@@ -415,9 +415,14 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
-          <groupId>org.collectionspace.services</groupId>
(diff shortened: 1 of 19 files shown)


ray-lee (approved):
Looks good.

ray-lee (approved):

PR #409, opened 2024-05-14
DRYD-1330: Add NAGPRA Claim

mikejritter:
**What does this do?**
Creates NAGPRA Claim service and id generator

Jira: https://collectionspace.atlassian.net/browse/DRYD-1330

This is the new claim procedure for the 8.1 release which will better support NAGPRA workflows.

* Rebuild collectionspace
* Start the tomcat server for collectionspace
* Run integration tests against the running collectionspace

This actually still needs tests to be added before merging.

No

@mikejritter built collectionspace and deployed to a local install

### DIFF
--- a/services/nagpraclaim/service/src/main/java/org/collectionspace/services/nagpraclaim/nuxeo/NagpraClaimValidatorHandler.java
+++ b/services/nagpraclaim/service/src/main/java/org/collectionspace/services/nagpraclaim/nuxeo/NagpraClaimValidatorHandler.java
@@ -1,0 +1,29 @@
+/*
+ * This document is a part of the source code and related artifacts
+ * for CollectionSpace, an open source collections management system
+ * for museums and related institutions:
+ *
+ * http://www.collectionspace.org
+ * http://wiki.collectionspace.org
+ *
+ * Licensed under the Educational Community License (ECL), Version 2.0.
+ * You may not use this file except in compliance with this License.
+ *
+ * You may obtain a copy of the ECL 2.0 License at
+ * https://source.collectionspace.org/collection-space/LICENSE.txt
+ */
+package org.collectionspace.services.nagpraclaim.nuxeo;
+
+import org.collectionspace.services.common.context.ServiceContext;
+import org.collectionspace.services.common.document.DocumentHandler.Action;
+import org.collectionspace.services.common.document.InvalidDocumentException;
+import org.collectionspace.services.common.document.ValidatorHandler;
+
+public class NagpraClaimValidatorHandler implements ValidatorHandler {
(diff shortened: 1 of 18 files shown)


services/nagpraclaim/service/src/main/java/org/collectionspace/services/nagpraclaim/nuxeo/NagpraClaimValidatorHandler.java
```diff
@@ -0,0 +1,29 @@
...
+import org.collectionspace.services.common.document.DocumentHandler.Action;
+import org.collectionspace.services.common.document.InvalidDocumentException;
+import org.collectionspace.services.common.document.ValidatorHandler;
+
+public class NagpraClaimValidatorHandler implements ValidatorHandler {
+
+    @Override
+    public void validate(Action action, ServiceContext ctx) throws InvalidDocumentException {
+        // TODO Auto-generated method stub
+        System.out.println("NagpraClaimValidatorHandler executed.");
```
ray-lee, line 27: Is there a reason we'd want make `claimNumber` required in records created through UI, but not the REST API? If not, there should probably be validation that `claimNumber` is not empty. See CollectionObjectValidatorHandler.

mikejritter: @ray-lee I updated the validation handler to check for the common part and claimNumber. I skipped using `CS_ASSERT` like the other validation handlers to avoid throwing an AssertionError just to rethrow the InvalidDocumentException. I did a quick test with the following json: 
```
{"document":{"ns2:nagpraclaims_common":{"@xmlns:ns2":"http://collectionspace.org/services/nagpraclaim","claimNumber":""}}}
```

Also do we want to update the other validation handlers so that they're making these two assertions? I know the authorities all implement theirs but I believe it's more sparse for procedures. I can make the ticket for this

ray-lee: Yes, I think we should have that ticket. I probably should have noticed this on PRs for other record types.

mikejritter: I should have asked as well, it's something I had noticed as being odd when creating procedures.

Ticket for the additional work is here: https://collectionspace.atlassian.net/browse/DRYD-1427

ray-lee (approved):
Looks good!

PR #410, opened 2024-05-24
DRYD-1331: NAGPRA Inventory

mikejritter:
**What does this do?**
* Create nagprainventory resource 
* Create nagprainventory idgenerator

Jira: https://collectionspace.atlassian.net/browse/DRYD-1331

* Build collectionspace with the application pr for nagpra inventory and all tenants needed for running the integration tests
* Start collectionspace
* Run the integration test suite

There are a few differences compared to normal, though nothing that prevents us from merging:

* client
  * pom: I've been trying to clean up the poms as much as I can by removing unneeded dependencies. As far as I can tell everything still works (api calls from the ui all succeed), and changes can be seen in 3cf92d5.
* jaxrs
  * xsd: Used the jaxb:class binding to set the class name, as noted below.
  * Previously there were some interfaces (JAXBSchema and ListItemJAXBSchema), which were unused so I opted to removed them. Change can be seen in 4881f6c.

No

@mikejritter tested locally


services/nagprainventory/client/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+    <dependency>
+      <groupId>org.collectionspace.services</groupId>
+      <artifactId>org.collectionspace.services.nagprainventory.jaxb</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.collectionspace.services</groupId>
+      <artifactId>org.collectionspace.services.person.client</artifactId>
+      <version>${project.version}</version>
+    </dependency>
```
mikejritter, line 47: Not sure why this dependency is included here, going to do a quick test without it.

services/nagprainventory/jaxb/src/main/resources/nagprainventory_common.xsd
```diff
@@ -0,0 +1,185 @@
...
+    See http://weblogs.java.net/blog/kohsuke/archive/2006/03/why_does_jaxb_p.html
+-->
+
+  <!--  NagpraInventory Information Group -->
+  <xs:element name="nagprainventories_common">
+    <xs:annotation>
+      <xs:appinfo>
+        <jaxb:class name="NagpraInventoriesCommon" />
+      </xs:appinfo>
+    </xs:annotation>
```
mikejritter, line 32: @ray-lee I wanted to highlight this as something new I was trying and is different from the other procedures. It just helps to generate a class name that's more consistent with what we'd expect, while still keeping the `napgrainventories_common` namespace in tact.

ray-lee (approved):
Looks good.

PR #411, opened 2024-06-06
DRYD-1333: Summary Documentation

mikejritter:
**What does this do?**
* Create summary documentation service
* Add id generator for summary documentation
* Create nagpra-jaxb module to hold common xsd elements

Jira: https://collectionspace.atlassian.net/browse/DRYD-1333

This adds the summary documentation procedure which is a necessary part of nagpra workflows.

* Rebuild collectionspace and run the integration tests
* Ideally we should also be checking that once deployed the summary documentation api responds correctly, e.g.
  * `curl http://localhost:8180/cspace-services/summarydocumentations`
  * also PUT with a payload to test the api, though the ui will take care of that as well

Some additional changes to the xsd portion of things which I did in kind of a hacky manner (normally it seems like you would use a catalog file to overwrite the schema location for the import but I did it directly). I moved the common data structures into `nagpra_common.xsd`, which is then imported in the jaxb module for the schemas to use. Not entirely necessary but something I had done to see what we can do with the schemas.

No

@mikejritter tested the api through the cspace-ui changes

### DIFF
--- a/services/JaxRsServiceProvider/pom.xml
+++ b/services/JaxRsServiceProvider/pom.xml
@@ -436,6 +436,12 @@
         <artifactId>org.collectionspace.services.nagprainventory.service</artifactId>
         <version>${project.version}</version>
       </dependency>
+        <dependency>
+            <groupId>org.collectionspace.services</groupId>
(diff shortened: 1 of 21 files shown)


ray-lee (approved):
Looks good.

PR #413, opened 2024-06-21
Convert Statements to PreparedStatements

mikejritter:
**What does this do?**
* Updates use of Statements to be PreparedStatements
* Wraps autocloseable classes in try-with-resources
* Adds cultural care fields to Held-in-Trust xsd

Jira (for Held-in-Trust): https://collectionspace.atlassian.net/browse/DRYD-1358

I had been playing around with some static analyses tools and they noted the use of Statements in the IDService. Since these queries are accessible through the API it seemed like a good time to go ahead and update them to be PreparedStatements instead, even if it just uses the path parameter for the csid value. I also opted to switch to try-with-resources as a way to handle closing of Connections/PreparedStatements/ResultSets to reduce some of the noise in the finally blocks.

* Rebuild collectionspace 
* Run the integration tests

This could be turned into 2 PRs just to separate the tasks. Initially I thought there might be more for the Held-in-Trust updates but the xsd changes were all that was needed. 

No

@mikejritter ran the integration tests and used this when testing the Held-in-Trust updates


ray-lee (approved):
Looks good. Thanks for doing those statement changes.

PR #416, opened 2024-07-15
DRYD-1399: RestrictedMedia Service

mikejritter:
**What does this do?**
* Add a RestrictedMedia client and service
* Add validator handler for RestrictedMedia
* Add document model handler for RestrictedMedia
* Add restricted media id generator
* Add restrictedmedia xsd

Jira: https://collectionspace.atlassian.net/browse/DRYD-1399

This adds a new type of Media procedure, RestrictedMedia, which is intended for use in workflows where there may be sensitive media which should not be immediately accessible. Currently it functions the same as the Media service, with the same logic for the Resource, Client, Proxy, and DocumentHandler classes. 

* Re-build and deploy collectionspace
* Start collectionspace
* Test that the restrictedmedia service is accessible, e.g.
```
curl -u 'admin@anthro.collectionspace.org:Administrator' http://localhost:8180/cspace-services/restrictedmedia/
```
* Run the integration tests

`RestrictedMediaResource#createBlob` is deprecated and likely doesn't make sense to carry over to a new service. It seems like the original deprecation (in MediaResource) was just in reference to using `PUT` vs `POST`.

I had been considering doing all the Blob handling for RestrictedMedia in the new Resource as well, but I opted to keep this simple for now as getting things setup to access the persistence layer was not as straightforward as I was expecting.

No

@mikejritter tested locally with the integration tests


mikejritter:
Closing this as we want to look at implementing restricted procedures in a more general way.

PR #419, opened 2024-07-29
Duty of Care Updates

mikejritter:
**What does this do?**
* Fix pluralization for duty of care
* Change title to dutyOfCareTitle
* Implement validation handler

Jira: https://collectionspace.atlassian.net/browse/DRYD-1332

These changes come from the testing done against Duty of Care. The pluralization update which was requested is being applied here for consistency throughout the procedure. I also noticed that the `title` field exists in the dublin core schema, which was sometimes being pulled back in the search results, so I've updated the field definition to be `dutyOfCareTitle` to avoid that collision.

* Test that the Duty of Care service is accessible on the new uri
```
curl --user 'admin@core.collectionspace.org:Administrator' http://localhost:8180/cspace-services/dutiesofcare
```

None

No

### DIFF
--- a/services/dutyofcare/client/src/main/java/org/collectionspace/services/client/DutyofcareClient.java
+++ b/services/dutyofcare/client/src/main/java/org/collectionspace/services/client/DutyofcareClient.java
@@ -14,14 +14,14 @@
  */
 package org.collectionspace.services.client;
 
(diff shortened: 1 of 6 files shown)


ray-lee (approved):
Looks good.

PR #420, opened 2024-07-29
Update basic object with current location report output type

mikejritter:
**What does this do?**
* Updates mime type for Basic Object with Computed Location

Jira: https://collectionspace.atlassian.net/browse/DRYD-1451

This is updates so that xlsx output is set by default, as it has a way to both provide tabular data and thumbnails. As the workflow is still being tested, we're only adding this to a single report so that we can understand how it might work with, e.g. mail merge.

* Rebuild collectionspace
* Create a collectionobject
* Select the Basic object with current location report and see xlsx as the default output 

None

No

@mikejritter tested locally


kspurgin:
I ended up here and was somehow able to convert it to a draft. I can change it back if that's annoying.

PR #421, opened 2024-08-28
DRYD-1451: Xlsx Report Configuration

mikejritter:
**What does this do?**
* Add xlsx configuration to jasper exporter
* Remove xlsx configuration properties from reports
* Refactor jasper exporter configuration to use newer exporters
* Update various logging messages to be more clear
* Wrap connection in try with resources
* Fix typos and some whitespace formatting

Jira: https://collectionspace.atlassian.net/browse/DRYD-1451

This is a follow up on testing exporting reports as xlsx files in order to display thumbnails. Instead of adding the headers to each individual report, we're adding them to the config directly and using that with the xlsx exporter. The headers have been removed from reports which previously had them as they're no longer necessary. I left the two `exclude` headers in, which from my testing didn't seem to change the report output.

I also did a bit of refactoring, mostly in the creation of the `Exporter` object. When changing from the parameter based config, the ExporterOutput had to match what each exporter expected. The easiest way to do this was to not use the generic `Exporter`, but each specific exporter which has the correct output type.

Other minor refactoring changes were just updating to try-with-resources and fixing some typos. I was considering doing more but didn't want to create too large of a PR.

* Create a few Collection Objects
* Create a few Media Handling records associated to the CollectionObjects
* Search for CollectionObjects in order to run the `Basic Object with Current Location` report
* Run the report with xlsx output (should be default now)
* Check the spreadsheet is formatted correctly and displays thumbnails

I think there could still be better error handling done here if we wanted to have more granular logging of what exactly failed. e.g. all the jasper errors get logged the same but could be differentiated (though the stack trace will at least do that).

No

@mikejritter tested locally using the `Basic Object with Current Location` report

### DIFF
--- a/services/report/3rdparty/jasper-cs-report/src/main/resources/accessions.xml
+++ b/services/report/3rdparty/jasper-cs-report/src/main/resources/accessions.xml
@@ -11,6 +11,6 @@
     <supportsGroup>false</supportsGroup>
     <supportsNoContext>true</supportsNoContext>
     <filename>accessions.jrxml</filename>
-    <outputMIME>text/csv</outputMIME>
+    <outputMIME>application/vnd.openxmlformats-officedocument.spreadsheetml.sheet</outputMIME>
   </ns2:reports_common>
 </document>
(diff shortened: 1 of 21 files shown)


ray-lee (approved):
Looks good!

PR #423, opened 2024-09-16
DRYD-1393: ServiceGroup filter on service tags 

mikejritter:
**What does this do?**
* Add tag element to service binding xsd
* Update ServiceGroup to filter on tag query parameter

Jira: https://collectionspace.atlassian.net/browse/DRYD-1393

These are changes that allow us to use a `tag` query on a subset of procedures, normally when searching for relations. The use case for this is the NAGPRA procedures which we want to create a sidebar for, and will do so by updating the base procedures in the application layer to have an additional `nagpra` tag. In the services layer the service binding stores those tags as a List, which is used when querying against the ServiceGroup `/items/` endpoint using a `tag` query parameter.

The general flow for filtering the doc types is as follows:
* If no `tag` parameter is found, search on all document types
* If a `tag` parameter is found, check that the service binding has the tag
* If no document types are available to query on after filtering, throw a 400

* Rebuild collectionspace with the associated application PR
* Create a few procedures
* Query against the service group endpoint and verify the correct procedures are returned, e.g.
```
# Expect all procedures to be returned
curl --user admin@core.collectionspace.org:Administrator http://localhost:8180/cspace-services/servicegroups/procedure/items?pgNum=0&pgSz=5&wf_deleted=false

# Expect nagpra procedures to be returned
curl --user admin@core.collectionspace.org:Administrator http://localhost:8180/cspace-services/servicegroups/procedure/items?pgNum=0&pgSz=5&wf_deleted=false&tag=nagpra

# Expect a 400 to be returned
curl --user admin@core.collectionspace.org:Administrator http://localhost:8180/cspace-services/servicegroups/procedure/items?pgNum=0&pgSz=5&wf_deleted=false&tag=notarealtag
```

I'm thinking about updating the error response still. Mostly to check that the `tag` query parameter was actually set (which should be the case but you never know).

I wanted to see about updating an integration test to check this as well. Probably in the services group module but I haven't had a chance to look around yet.

Also the `/cspace-services/servicegroups/procedure` endpoint still needs to be updated for the tags as well. Not sure if we want to handle that here or make a new ticket and update it when we add sections to the ui. I thought a ticket for the ui changes already existed but I can't seem to find it atm (maybe DRYD-1346 but I thought there were some mock ups, might be misremembering).

No

@mikejritter tested locally


ray-lee:
@mikejritter We'll need to be able to exclude tags, e.g. for a sidebar component that shows related non-NAGPRA records. Maybe we could keep the `tag` parameter that you have, but allow exclusions preceded by `-`. So `tag=-nagpra` would be all record types not tagged `nagpra`, `tag=ownership,-nagpra` would be all record types tagged `ownership` that are not also tagged `nagpra`. What do you think?

ray-lee:
Maybe it would be `tag=ownership&tag=-nagpra`. However the API is expecting multiple values for a parameter.

mikejritter:
@ray-lee That's along the lines of what I was thinking too for excluding tags. The only other option I can think of would be adding another query parameter, which is kind of awkward but would work mostly the same.

At the moment I was restricting it to a single parameter (comma separated would fail to match any and return a 400; additional params are ignored), but there's no reason we can't do multiple values. The only question I think we would need to answer is if it should match against all or any of the provided tags. From your example it sounds like we'd be matching against all.

mikejritter:
@ray-lee I pushed a few commits to allow for tags to be excluded, let me know if it's along the lines of what you were thinking for it. I had one remaining question around empty strings which I still need to verify.

I also added some unit tests. I realize I meant to use TestNG for it but ended up doing JUnit by default. They're pretty simple tests though, just to make sure the little bit of logic being added is correct.

ray-lee:
@mikejritter I'm now thinking we should hold off on accepting multiple tags, since we don't need it right now. There are multiple ways of interpreting what it means to have multiple includes and excludes, and I'm not sure which one we want, so it's probably better to wait until we have an actual use case.

ray-lee:
I'd rather use testng to be consistent. It will start to be confusing if we're using both.

mikejritter:
@ray-lee I pushed a few commits to check for only a single query parameter and use testng. I still want to verify that everything is working via the api but I imagine it should be fine.

services/client/src/main/java/org/collectionspace/services/client/IQueryManager.java
```diff
@@ -51,6 +51,7 @@ public interface IQueryManager {
 	final static String SEARCH_QUALIFIER_OR = SEARCH_TERM_SEPARATOR + "OR" + SEARCH_TERM_SEPARATOR;
 	final static String DEFAULT_SELECT_CLAUSE = "SELECT * FROM ";
 	final static String CSID_QUERY_PARAM = "csid";
+	String TAG_QUERY_PARAM = "tag";
```
ray-lee, line 54: Should put `final static` here, like the other constants. (The conventional order is `static final`, but  🤷)

I'm thinking we should name the query parameter `servicetag`, in case we add tags to individual records or something in the future. It's a bit unclear what "tag" applies to when you look at the URL. Using `tag` in the service and application config is fine, since it's clear in context what's being tagged.

services/servicegroup/service/src/main/java/org/collectionspace/services/servicegroup/nuxeo/ServiceGroupDocumentModelHandler.java
```diff
@@ -162,79 +160,138 @@ public PoxPayloadOut getResourceItemForCsid(ServiceContext<PoxPayloadIn, PoxPayl
...
+	 * does not contain said tag.
+	 *
+	 * @param binding the ServiceBinding to check
+	 * @param queryTag tag query parameter
+	 * @return true if the ServiceBinding contains all query parameters, false otherwise
+	 */
+	public boolean acceptServiceBinding(ServiceBindingType binding, String queryTag) {
+		final Tags tags = binding.getTags();
+		if (tags == null || tags.getTag().isEmpty()) {
+			return false;
```
ray-lee, line 272: Can't return false here now that we have negation. You probably just want to initialize null to an empty list, and let it fall through.

mikejritter: Oh yep, good catch. I'll update this and get a test in for it too.

mikejritter:
@ray-lee I've been meaning to ask - do we want to handle the updates for the `/cspace-services/servicegroups/procedure` endpoint returning the tags for each record type here as well? Otherwise we can get it done with the work for https://collectionspace.atlassian.net/browse/DRYD-1351.

ray-lee:
@mikejritter I think it makes sense to do it with DRYD-1351.

ray-lee (approved):
Looks good!

PR #424, opened 2024-09-17
DRYD-1355: Deaccession Service

mikejritter:
**What does this do?**
* Creates a deaccession service
* Creates a deaccession id_generator

Jira: https://collectionspace.atlassian.net/browse/DRYD-1355

The Deaccession procedure adds the ability to capture the decision making around removing an accessioned object. Previously this was part of the ObjectExit procedure, and this is breaking it out into it's own data structure.

* Rebuild and start collectionspace with all tenants required for integration tests
* Run the integration tests and ensure they pass

None

No

@mikejritter tested locally

### DIFF
--- a/services/common/src/main/resources/db/postgresql/load_id_generators.sql
+++ b/services/common/src/main/resources/db/postgresql/load_id_generators.sql
@@ -1270,3 +1270,39 @@ INSERT INTO id_generators
         SELECT  csid
         FROM    id_generators
         );
+
+-- DEACCESSION_NUMBER
+
+INSERT INTO id_generators
+    (csid, displayname, description, priority, last_generated_id, id_generator_state)
+  SELECT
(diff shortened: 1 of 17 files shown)


services/common/src/main/resources/db/postgresql/load_id_generators.sql
```diff
@@ -1270,3 +1270,39 @@ INSERT INTO id_generators
...
         );
+
+-- DEACCESSION_NUMBER
+
+INSERT INTO id_generators
+    (csid, displayname, description, priority, last_generated_id, id_generator_state)
+  SELECT
+     '0b508d05-67df-4b63-b4ce-7319a3029baa',
+     'Deaccession Number',
+     'Identifies a consultation document.',
```
ray-lee, line 1281: "consultation" copy/paste error?

ray-lee (approved):
Looks good.

PR #426, opened 2024-10-04
Support for deurning arrays in reports

mikejritter:
**What does this do?**
* Add deurn function for arrays

This is to support new reports [DRYD-1334](https://collectionspace.atlassian.net/browse/DRYD-1334) and [DRYD-1335](https://collectionspace.atlassian.net/browse/DRYD-1334). It allows us to deurn in the services layer rather than the database. 

* Pass an array in to the deurn fields
* See that it is deurned

@ray-lee I wanted to get this in first before doing the reports as I have a few thoughts/concerns on it first.

I opted to set the field to a List, but I'm still debating this. This is fine for groovy, but when using Rhino for the jasper expressions there isn't really good support for Lists. I think instead we probably should just do an array which seems like it will work better for both.

Likewise the jrxml has some pain points as well. When defining the class for the field, java.sql.Array is normally used, which makes sense. However, to create a new sql Array you need the db connection then you can assign the primitive array to it. I opted not to do this and instead just set the field type to java.util.List. It works, but is kind of quirky and was something I wanted to bring up. If you need an example of it lmk and I'll add some samples.

No

@mikejritter used these for the new reports

### DIFF
--- a/services/report/service/src/main/java/org/collectionspace/services/report/jasperreports/CSpaceReportScriptlet.java
+++ b/services/report/service/src/main/java/org/collectionspace/services/report/jasperreports/CSpaceReportScriptlet.java
@@ -1,6 +1,10 @@
 package org.collectionspace.services.report.jasperreports;
 
+import java.sql.Array;
+import java.sql.SQLException;
+import java.util.ArrayList;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 
 import net.sf.jasperreports.engine.JRDefaultScriptlet;
@@ -75,8 +79,14 @@ public class CSpaceReportScriptlet extends JRDefaultScriptlet {
 					continue;
 				}
 
+				// kind of janky - we want the correct type in the jrxml (java.util.List), so we can't check the
+				// value class. Instead, get the value which should be a java.sql.Array.
+				if (field.getValue() instanceof Array) {
+					deurnArray(field);
+				}
+
 				if (!field.getValueClass().equals(String.class)) {
-					logger.warn("{}: deurn field is not a string: {}", getReportName(), fieldName);
+					logger.warn("{}: deurn field is not a string or array: {}", getReportName(), fieldName);
 
 					continue;
 				}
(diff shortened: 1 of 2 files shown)


ray-lee:
@mikejritter Could you point me to some example jrxml that uses this? I think that will help me understand the concerns.

mikejritter:
@ray-lee it turns out I forgot to push my other branch with the reports, so I only have a brief example for now:
```
...
<!-- pass in an array to the deurn fields -->
	<parameter name="deurnfields" class="java.lang.String" isForPrompting="false">
		<defaultValueExpression><![CDATA[cultures]]></defaultValueExpression>
	</parameter>
<!-- here we set the class type for jasper - since the deurn returns a list that's what we end up using -->
	<field name="cultures" class="java.util.List">
		<property name="com.jaspersoft.studio.field.label" value="cultures"/>
		<property name="com.jaspersoft.studio.field.tree.path" value="culture_group"/>
	</field>

<!-- and an example of a normal array - this is what jasper actually gets back from the query -->	
<field name="acquisition_numbers" class="java.sql.Array">
		<property name="com.jaspersoft.studio.field.label" value="acquisiition_numbers"/>
		<property name="com.jaspersoft.studio.field.tree.path" value="acquisitions_common"/>
	</field>
...
<!-- these would be the example detail fields, where we get extra operations if we're using groovy (this example) or js -->
			<textField isStretchWithOverflow="true">
				<reportElement x="540" y="0" width="50" height="30" uuid="ec56afe3-43e1-498f-9c80-b969fce26489">
					<property name="com.jaspersoft.studio.spreadsheet.connectionID" value="7ecd3a48-36c0-40b7-8af5-001e23c80705"/>
				</reportElement>
				<textFieldExpression><![CDATA[$F{culture}.join(';')]]></textFieldExpression>
			</textField>
			<textField isStretchWithOverflow="true">
				<reportElement x="590" y="0" width="50" height="30" uuid="e7a69d4f-c29e-4e36-b19d-8425e309e063">
					<property name="com.jaspersoft.studio.spreadsheet.connectionID" value="2a426f37-edf0-42c8-9b7d-70f8b83cad0b"/>
				</reportElement>
				<textFieldExpression><![CDATA[$F{acquisition_numbers}.getArray().join(';')]]></textFieldExpression>
			</textField>

```

I think the biggest issue to me is that Rhino doesn't seem to support Lists, so the text expression for `cultures` won't work when using js. Though I think you could do something like `cultures.toArray().join(';')` and get it to work. Now that I've had a chance to sit on it a little I don't think it's _that_ bad of a thing, it's just a little awkward.

ray-lee:
@mikejritter I think what I would do is to create our own class that implements `java.sql.Array`, called something like `DeurnedArray`. It would wrap the java.sql.Array returned by the query. `DeurnedArray` would pass all its methods through to the wrapped java.sql.Array, except that the `getArray(...)` methods would deurn the values returned by the wrapped method, before returning them to the caller. What do you think?

mikejritter:
@ray-lee I just updated with the new class to provide the Array interface. I opted only to include the `getArray()` implementation - poking around it doesn't seem like they're even fully supported from the jdbc postgres driver so I think it should be ok. Plus this is for our own use so I don't expect we'll need the slices, which could be added in the future. 

I think this is in a good spot now. Null arrays can still be problematic (jasper doesn't print any errors when running the text expressions), but that's a separate issue probably better handled by the reports.

ray-lee (approved):
Looks good.

PR #433, opened 2024-11-20
DRYD-1351: API For Supporting Create Page Update

mikejritter:
**What does this do?**
* Add new endpoint /servicegroups/:group/tags
* Add servicetag query param to /servicegroups/:group
* Apply spotless

Jira: https://collectionspace.atlassian.net/browse/DRYD-1351

We want to be able to organize procedures on the Create New page into groups. In order to do this we added support for tagging procedures, but didn't update the API to support what tags exist and displaying record types which have a given tag. These changes add both of those to the ServiceGroupResource which allows us to finish the workflow in the ui.

* Test the new `/servicegroups/:group/tags` endpoint
  * e.g. `curl -H "Accept: application/json" -v --user admin@core.collectionspace.org:Administrator http://localhost:8180/cspace-services/servicegroups/procedure/tags`
  * `curl -H "Accept: application/json" -v --user admin@core.collectionspace.org:Administrator http://localhost:8180/cspace-services/servicegroups/authority/tags`
* Test the query parameter on `/servicegroups/:group`
  * `curl -H "Accept: application/json" -v --user admin@core.collectionspace.org:Administrator "http://localhost:8180/cspace-services/servicegroups/procedure?servicetag=nagpra"`

This lacks tests, though some could be added to the `ServiceGroupServiceTest`.

No. API documentation will need to be updated for the new endpoint and query parameter.

@mikejritter tested locally

### DIFF
--- a/services/servicegroup/jaxb/src/main/resources/servicegroup_common.xsd
+++ b/services/servicegroup/jaxb/src/main/resources/servicegroup_common.xsd
@@ -11,34 +11,31 @@
     $LastChangedDate: 2010-06-02 16:03:51 -0700 (Wed, 02 Jun 2010) $
 -->
 
-<xs:schema 
+<xs:schema
   xmlns:xs="http://www.w3.org/2001/XMLSchema"
   xmlns:jaxb="http://java.sun.com/xml/ns/jaxb"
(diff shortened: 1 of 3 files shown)


axb21:
@mikejritter I tested this code locally and the endpoints look good.

PR #437, opened 2024-12-11
Add relatedLinks, featuredCollection displayName to elastic index

mikejritter:
**What does this do?**
* Index new related link fields
* Index displayName for featuredCollection

Jira:
* https://collectionspace.atlassian.net/browse/DRYD-1455
* https://collectionspace.atlassian.net/browse/DRYD-1456

The related link group list is a new field for 8.1 which is meant for the public browser. This adds the field to the elastic index so it can be... indexed.

Likewise there was a request to have the Featured collection be clickable with the link bringing you to the search results on the featured collection. Since the filtering is normally done using a human readable name, this adds the displayName field to the index for featuredCollections so that it may be queried on.

I need to do a write up for deploying ElasticSearch and the cspace-public-gateway to the wiki as configuration needs to be updated for both CollectionSpace and the gateway in order to index + retrieve. 

Currently there are no tests for elasticsearch - we could look into using TestContainers with their elasticsearch support in the future.

None

No; new indexes should be documented

@mikejritter tested the `publishedRelatedLinkGroupList`, however I ran into issues when testing materials. My intent here is to use dev to test if the fields are being indexed correctly.

### DIFF
--- a/services/common/src/main/cspace/config/services/tenants/fcart/fcart-tenant-bindings.delta.xml
+++ b/services/common/src/main/cspace/config/services/tenants/fcart/fcart-tenant-bindings.delta.xml
@@ -53,6 +53,7 @@
 							"collectionobjects_common:objectStatusList",
 							"collectionobjects_common:otherNumberList",
 							"collectionobjects_common:ownersContributionNote",
+							"collectionobjects_common:publishedRelatedLinkGroupList",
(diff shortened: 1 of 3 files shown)


mikejritter (commented):
I was able to build and deploy CollectionSpace with these changes, so I'm going to merge them in order for them to be deployed from the nightly build.

PR #438, opened 2024-12-12
8.1 Report QA Changes

mikejritter:
**What does this do?**
* DRYD-1431: Update descriptions for reports
* Exhibition checklist - deurn computedcurrentlocation
* Exhibition checklist - Add dimension measured part
* Inventory consultation - Increase height of header bar
* Inventory consultation - Add inventory date to status
* Notice of Intent to Repatriate - Switch to javascript templating
* Notice of Intent to Repatriate - Pull distinct acquisitions related to summary
* Notice of Inventory Completion - Switch to javascript templating
* Notice of Inventory Completion - Correct acquisition date join
* Notice of Inventory Completion - Correct object count joins
* Repatriation consultation - Increase height of header bar
* Misc - Use LocalDateTime in an attempt to avoid timezone issues
* Misc - Use coalesce to avoid null/zero values
* Misc - Filter null values for dates/lists

Jira:
* https://collectionspace.atlassian.net/browse/DRYD-1431 
* https://collectionspace.atlassian.net/browse/DRYD-1505
* https://collectionspace.atlassian.net/browse/DRYD-1329

These are updates from QA on the new reports for 8.1. A good chunk are updates to better handle null values by using coalesce or filtering them when in the template. There are also some issues with formatting which are still being combed through - header heights have increased in a few areas and the spacing/width of the page has been adjusted for some reports.

There were a few missing fields (dimension measured part, inventory date) which have been added. Some fields were also missing because the joins were not done correctly (object count was using the csid instead of the id; acquisition date had a typo in the name). 

We also saw that when deploying to the dev servers dates were being reported incorrectly. I've switched to LocalDateTime to try and match what comes from the database better as the timezone isn't stored.

* Run the updated reports and verify changes are in place

None

No


mikejritter:
I split this into multiple PRs for each report/report grouping.

PR #439, opened 2024-12-13
Report: Notice of Intent to Repatriate QA

mikejritter:
**What does this do?**
* Switch to javascript templating
* Pull distinct acquisitions related to summary
* Coalesce for better null handling
* Adjust spacing for some fields

Jira: https://collectionspace.atlassian.net/browse/DRYD-1334

These are changes that came out of initial testing for the Notice of Intent to Repatriate report. Notably we had duplicating values; we now use distinct on the summary id for acquisitions to try and limit this. There will likely be some repeated data if multiple acquisitions have the same values but that is something which is still up for discussion. Likewise we are now using coalesce in order to prevent null arrays, and using filter when joining array values in order to remove null entries.

* Rebuild collectionspace
* Run the report and verify all changes

None

No

@mikejritter tested locally


mikejritter (commented):
For the filters we really should wrap the parameters with parens. This follows the eslint recommendations; it's kind of tricky to test against these embedded scripts though.

PR #440, opened 2024-12-13
DRYD-1431: Add descriptions to reports

mikejritter:
**What does this do?**
* Adds descriptions to reports

Jira: https://collectionspace.atlassian.net/browse/DRYD-1431

This adds descriptions to reports which had placeholders. Note that this is only for older reports, not newer ones for 8.1.

* Build a clean install for CollectionSpace
* See that the reports have appropriate descriptions

None

No

Also no.


mikejritter:
Adding a note that before merging I'm considering adding the fields to the descriptions.

PR #451, opened 2025-02-11
DRYD-1660: Aggregate collections for objects

mikejritter:
**What does this do?**
* Update to aggregate all `collectionobjects_publicart_publicartcollections` for a collection object

Jira: https://collectionspace.atlassian.net/browse/DRYD-1662

This was a request which came in from Creative West for an update to the report. I opted to update the scripting language to js because the fields to be de-urned and we now have support for de-urning arrays. So we can join the array rather than the regex + string agg. I'm not actually sure which method is better but trying to be consistent with something.

* Rebuild and start collectionspace with the publicart profile enabled
* In the publicart profile
  * Create a collectionobject
  * Assign multiple collections to the object
  * Save the object
  * In the report tab, run the `Full Object with Place Details` report
  * See each collection separated by a `;`

None

No

@mikejritter tested locally with multiple and empty collections


mikejritter:
This needs to be rebased on v8.1-branch before merge to prevent the 8.2 version bump from being included.

PR #452, opened 2025-02-18
DRYD-1683: SSO expects typed AttributeValues

axb21:
**What does this do?**
CollectionSpace was implicitly assuming that attribute values sent by an identity provider during SAML-based SSO logins were explicitly typed as strings. By contrast, the SAML 2.0 specifies that identity providers can send them typed as strings or untyped. Some CollectionSpace customers that want to use SAML SSO with CollectionSpace send untyped attributes. In this instance, logins would fail because CollectionSpace would not find attribute values such as user email addresses that were expected. This fix puts CollectionSpace's behavior more in line with the SAML specification and should be able to successfully find email address and other attribute values that are not explicitly typed.

This PR concerns https://collectionspace.atlassian.net/browse/DRYD-1683, DRYD-1683 (SSO expects typed AttributeValues).

This PR includes a new set of unit tests org.collectionspace.services.common.test.SecurityUtilsTest that test for correct extraction of attribute values at the unit level. One unit test was designed to fail with CollectionSpace's current behavior and succeed once the modification to attribute value extraction was modified to be in line with the SAML 2.0 specification.

This PR should also be tested in a running environment where the identity provider (e.g., Auth0) should be explicitly configured to send untyped attribute values.

None

I do not believe the documentation needs to be modified.

I ran the unit tests locally but did not try running CollectionSpace configured to use SAML SSO.

### DIFF
--- a/services/common/src/test/java/org/collectionspace/services/common/test/SecurityUtilsTest.java
+++ b/services/common/src/test/java/org/collectionspace/services/common/test/SecurityUtilsTest.java
@@ -1,0 +1,222 @@
+package org.collectionspace.services.common.test;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testng.Assert;
+import org.testng.annotations.BeforeSuite;
+import org.testng.annotations.Test;
+import org.w3c.dom.Element;
+
+import java.util.ArrayList;
+import java.util.Set;
+
+import javax.xml.namespace.QName;
+import javax.xml.transform.OutputKeys;
+import javax.xml.transform.Transformer;
+import javax.xml.transform.TransformerConfigurationException;
+import javax.xml.transform.TransformerException;
+import javax.xml.transform.TransformerFactory;
+import javax.xml.transform.dom.DOMSource;
+import javax.xml.transform.stream.StreamResult;
+
+import org.collectionspace.services.common.security.SecurityUtils;
+import org.collectionspace.services.config.AssertionAttributeProbeType;
+import org.collectionspace.services.config.AssertionProbesType;
+import org.joda.time.DateTime;
+import org.opensaml.core.config.ConfigurationService;
+import org.opensaml.core.config.InitializationException;
+import org.opensaml.core.config.InitializationService;
+import org.opensaml.core.xml.XMLObject;
+import org.opensaml.core.xml.XMLObjectBuilder;
+import org.opensaml.core.xml.XMLObjectBuilderFactory;
+import org.opensaml.core.xml.config.XMLObjectProviderRegistry;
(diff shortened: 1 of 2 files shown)


services/common/src/test/java/org/collectionspace/services/common/test/SecurityUtilsTest.java
```diff
@@ -0,0 +1,222 @@
...
+    public void assertionWithUntypedAttributeValuesIsNotNull() {
+    	testBanner("the mock assertion with untyped attribute values is not null");
+    	Assert.assertNotNull(testAssertionUntypedAttributeValues);
+    }
+    @Test(dependsOnMethods = {"assertionWithTypedAttributeValuesIsNotNull"})
+    public void candidateUsernamesTypedNotNullOrEmpty() {
+    	testBanner("findSamlAssertionCandidateUsernames finds candidate usernames when they are typed as string");
+    	Set<String> candidateUsernames = SecurityUtils.findSamlAssertionCandidateUsernames(testAssertionTypedAttributeValues, null);
+		Assert.assertNotNull(candidateUsernames);
+		if(null != candidateUsernames)
```
mikejritter, line 211: Small nitpick but it seems like with the assertion above we don't need the if statement.

mikejritter (commented):
@axb21 Looks good, just one small thing I noticed.

I'm going to try to test this locally w/ auth0 since it was mentioned we might be able to have untyped attributes depending on the configuration.

axb21:
> @axb21 Looks good, just one small thing I noticed.

Do you mean the comment below about `assertNotNull` or was there something else?

> I'm going to try to test this locally w/ auth0 since it was mentioned we might be able to have untyped attributes depending on the configuration.

Great!

balmas:
@mikejritter take a look at Scenario 11 at https://collectionspace.atlassian.net/wiki/spaces/COL/pages/3253501955/SSO+Testing which contains the setting you need to update to switch to untyped attributes in Auth0

services/common/src/test/java/org/collectionspace/services/common/test/SecurityUtilsTest.java
mikejritter: Yea like I said it was just a nitpick and was something that I noticed. I hadn't thought of someone updating the tests in such a way that the previous assertion would be removed but I guess crazier things have happened :).

mikejritter (approved):
Created an additional test with some xml which didn't include the `xsi:type` attribute in order to confirm the assumption that the xml is unmarshalled into the XSAny type. I don't think it's something we would need to include so I'll just add some notes about it to the Jira ticket.

With that done I'm going to merge this in to the v8.1-branch.

PR #456, opened 2025-03-18
DRYD-1716: set the antrun plugin that runs check-env-vars to <inherited>false</inherited>

axb21:
**What does this do?**
The `maven-antrun-plugin` was running a goal `check-env-vars` once for every child project within the `services` multi-module project. Besides being a redundant check--the env cannot change while maven is executing--and besides being a deploy-only necessity, it also added many seconds to compile times. This PR changes the build process to only run `check-env-vars` once per compile cycle, saving ~5 seconds of compile time.

Addresses JIRA ticket [DRYD-1716](https://collectionspace.atlassian.net/browse/DRYD-1716)

This is a build-level change so there are no unit tests. It can be tested by running commonly-used maven commands and verifying they are behaving as expected.

N/A

N/A

@axb21 ran this code locally. `mvn compile`, `mvn test`, and `mvn clean install` all work as expected.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -385,6 +385,7 @@
 			<plugin>
 				<artifactId>maven-antrun-plugin</artifactId>
 				<version>1.8</version> <!-- Most recent version is 3.0.0 -->
+				<inherited>false</inherited>
 				<executions>
 					<execution>
 						<id>check-environment-vars</id>


axb21:
Sorry, I inadvertently closed this PR by prematurely deleting the branch. It's reopened now.

mikejritter:
@axb21 just fyi I might update the target branch to `main` because hosting is expecting the `v8.1-branch` to be immutable.

mikejritter (approved):
Works as advertised :+1: 

This does impact some of the ant tasks if they're run from their respective directories but we have a jira in for having them check their vars themselves so they'll be handled separately.

PR #457, opened 2025-03-18
DRYD-1712/DRYD-1715: only run `git-commit-id-plugin` once

axb21:
**What does this do?**
Our build had `git-commit-id-plugin` in the topmost `pom.xml` file from which all other `pom.xml` files inherit. This plugin' reads the latest git commit and stores it in a file `git.properties` that is read by the `systeminfo` service. See for instance; 
systeminfo-screenshot

this shows the version of CSpace services running comes from commit `a7af564` (the<build> tag).

However, because of this inheritance structure the `git-commit-id-plugin` was being run 217 times during builds, even builds that don't need this number. This unnecessary extra work added tens of seconds to build times. 

This change removes the `git-commit-id-plugin` from the topmost `pom.xml` and places it instead in the `JaxRsServiceProvider` project. The latter is where the ultimate WAR file used to deploy Cspace services is built, and maven can find the generated `git.properties` file there.

This PR fixes the bug described in [DRYD-1712](https://collectionspace.atlassian.net/browse/DRYD-1712). It also addreses the enhancement in [DRYD-1715](https://collectionspace.atlassian.net/browse/DRYD-1715), which is no longer needed since this plugin is not taking up significant time during the build now that this bug is fixed.

N/A

No merge dependencies. There should not be implications for production.

N/A

@axb21 ran this code locally. I packaged the WAR `cspace-services.war` file and verified `git.properties` was in it and had the intended contents. I also deployed CSpace services from this commit and verified that the `systeminfo` service runs correctly and reports the correct SHA (compare the value in the <build> tag reported by the `systeminfo` service with the 1st commit SHA in the console output of `git`): 
systeminfo-screenshot
console-screenshot

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -414,11 +414,6 @@
                     </files>
                 </configuration>
             </plugin>
-
-			<plugin>
-				<groupId>pl.project13.maven</groupId>
-				<artifactId>git-commit-id-plugin</artifactId>
-			</plugin>
 		</plugins>
 	</build>
 
--- a/pom.xml
+++ b/pom.xml
@@ -227,29 +227,6 @@
 						</archive>
 					</configuration>
 				</plugin>
-
-				<plugin>
-					<groupId>pl.project13.maven</groupId>
-					<artifactId>git-commit-id-plugin</artifactId>
-					<version>4.9.10</version>
-					<executions>
-						<execution>
-							<goals>
-								<goal>revision</goal>
-							</goals>
-							<phase>initialize</phase>
-						</execution>
-					</executions>
-					<configuration>
-						<generateGitPropertiesFile>true</generateGitPropertiesFile>
-						<generateGitPropertiesFilename>${project.build.outputDirectory}/git.properties</generateGitPropertiesFilename>
-						<includeOnlyProperties>
-							<includeOnlyProperty>^git.commit.id.abbrev$</includeOnlyProperty>
-						</includeOnlyProperties>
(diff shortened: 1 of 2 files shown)


mikejritter (changes requested):
@axb21 I noticed mvn is reporting warnings now with respect to the config of the `git-commit-id-plugin`, e.g.:
```
[WARNING] Some problems were encountered while building the effective model for org.collectionspace.services:org.collectionspace.services.3rdparty.nuxeo.collectionspace_core:jar:8.1.1
[WARNING] 'build.plugins.plugin.version' for pl.project13.maven:git-commit-id-plugin is missing. @ org.collectionspace.services:org.collectionspace.services:${revision}, /home/shake/cspace/services/pom.xml, line 418, column 12
```

It looks like this is moving it from the `pluginManagement` section in the parent pom but it's still defined in the `plugins` section below.

axb21:
> @axb21 I noticed mvn is reporting warnings now with respect to the config of the `git-commit-id-plugin`, e.g.:

Hmm that looks like a bug on my part. Let me double check this before we merge.

axb21:
> It looks like this is moving it from the `pluginManagement` section in the parent pom but it's still defined in the `plugins` section below.

@mikejritter Yes, that was it. I forgot to remove it from <plugins> in the topmost pom. I pushed a change to fix that and verified the WAR still packages correctly. I did not redeploy and retest the services since the semantics shouldn't be affected.

mikejritter (approved):

PR #459, opened 2025-03-31
DRYD-1742: Change deployment for common/lib jars

mikejritter:
**What does this do?**
This creates a new module `common-lib` which manages the binary jars dependencies through maven. It adds an ant target (`deploy`) which can be used to deploy the dependencies into the tomcat `lib` directory.

Jira: https://collectionspace.atlassian.net/browse/DRYD-1742

The jars which currently exist in the repository take up a lot of space in addition to being difficult to manage. By using maven, we're making it easier to track what versions we're deploying and what should be updated.

* In the `common-lib` module:
  * Without any environment variables set, try `ant deploy` and see that it fails
  * Set your `CSPACE_JEESERVER_HOME` and run `ant deploy` and see that the dependencies are copied into the `${CSPACE_JEESERVER_HOME}/lib` directory
* Run a full deploy and see that CollectionSpace starts and is usable via cspace-ui

There is still work which will be done for:
* Pushing JARs in `services/common/lib` into our maven repository (if needed)
* Removing the JARs in `services/common/lib` from the git repository
* Doing the same tasks for the jars in `3rdparty/nuxeo-server` and `3rdparty/jdbc_drivers`

An undeploy target is still needed to reset Tomcat to it's default state.

No. Documentation shouldn't need to be updated as the `any deploy` task will still be the main way to deploy CollectionSpace.

@mikejritter has tested deploy to a local CollectionSpace, is running integration tests/cspace-ui to verify everything else


axb21:
I'm cool with merging this.

PR #461, opened 2025-04-09
Dryd 1714 proposal

axb21:
**What does this do?**
The CollectionSpace `services` project uses JAXB to code generate Java classes from XML Schema Document definitions. We rarely change the XSD documents, yet without this PR many of the phases in the Maven build trigger re-generation of these Java classes. This is wasted work in almost all cases and adds many seconds to the build. Furthermore, it is clear when the Java classes should be regenerated: either on a clean checkout of the code/first time compilation; or, whenever one of the XSD files must be changed to add a feature or fix a bug. So, this PR adds a property that can be used to turn JAXB code generation on or off. By default it is on and the build behaves as usual; however, there is a new property `-Djaxb.codegen.phase=none` that, when set, runs the build without code generation. 

This PR address Jira ticket https://collectionspace.atlassian.net/browse/DRYD-1714 , "Create a maven profile in the `services` maven build for controlling whether JAXB code generation runs"

There are no unit tests since this is a build-level change. To test the build:
- Run `mvn PHASE` and verify PHASE runs as usual, including invocation of the Maven `maven-jaxb2-plugin`
- Run `mvn PHASE -Djaxb.codegen.phase=none` and verify that PHASE runs without invoking the `maven-jaxb2-plugin` goal(s).

N/A

N/A

@axb21 has run compile, test, package, and install build phases with and without the `-Djaxb.codegen.phase=none` property.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ bin
 *.diff
 logged_schemas
 logs
+.mvn
--- a/build.properties
+++ b/build.properties
@@ -9,7 +9,7 @@ host=127.0.0.1
 #
 # Release version info
 #
-release.version=8.1.1
+release.version=8.2.0-SNAPSHOT
 cspace.release=${release.version}
 cspace.instance.id=${env.CSPACE_INSTANCE_ID}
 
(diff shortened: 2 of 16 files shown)


mikejritter (approved):
Looked at this earlier and it seems simple, non-disruptive, and shaves a little bit of time off the build. :+1:

PR #462, opened 2025-04-09
DRYD-1754: Manage Nuxeo JARs through Ant/Maven

mikejritter:
**What does this do?**
This creates a pom and build xml files which server as a basis for deploying the nuxeo server components to Tomcat. This is done by adding the nuxeo modules necessary for running as an embedded server into the pom.xml and using the maven copy-dependency plugin in order to copy JARs directly into the `nuxeo-server/bundles` and `nuxeo-server/lib` directories. The `nuxeo-server/lib` directory contains dependencies of the nuxeo bundles directory and as such we do not manage any unless we need to override versions.

Jira: https://collectionspace.atlassian.net/browse/DRYD-1754

The current deployment strategy relies on the JARs to be embedded in the git repository. Not only are these dependencies difficult to manage, but they also take up considerable space in the repository. By switching to maven for managing the versions, we can have an easier time seeing what the transitive dependencies are and if they can/should be updated. 

I'm not sure if this will help in the future for upgrading Nuxeo, but it is a first step in moving to a new deployment strategy.

Test that the environment check works:
* cd to `3rdparty/nuxeo/nuxeo-server`
* run `ant deploy` without the CSPACE_JEESERVER_HOME set
* see that it fails

Test that the new deployment works:
* Set your environment for the CollectionSpace build
* Build CollectionSpace as usual
* See that the `${CSPACE_JEESERVER_HOME}/nuxeo-server/{bundles,lib}` directories are populated
* Interact with CollectionSpace through the cspace-ui.js project or the api

There are a number of changes to the deployed JARs which we should consider.

*Exclusions*
The first thing I want to bring up are the exclusions in the `nuxeo-server/pom.xml`. The `nuxeo-platform-audit-core` was excluded because when bringing it in it was looking for a datasource via jndi and logging errors when it could not be found. I'm sure there's a way to disable it but I didn't spend much time on it. 

The other exclusion to highlight is `jackson-core`, which we already bring in and don't want to re-deploy. The version of nuxeo we're on brings in an older version which causes NoClassDefFound issues, and if we were to deploy the same version we can get LinkageErrors because the jar exists in multiple places causing classes to load multiple times.

*Bundle JAR Updates*
There are ~31 new JARs which are brought in as transitive dependencies. It's possible that some of the `-core` JARs are running additional services via osgi similar to what happened with the audit dependency. If we want a more one-to-one deployment, we can add exclusions to remove dependencies.

Only 2 dependencies were removed - one for redis and one mongodb. We don't deploy either so I didn't see a reason to keep them around.

*Lib JAR Updates*
The `lib` folder has a significant number of JARs removed. Most seemed to be related to modules which we weren't using (e.g. directory-ldap, redis, etc) and at a glance things are working without them (I tested on a normal deployment with them removed).

Added dependencies are transitive dependencies from nuxeo modules we bring in. I'm not sure how many come from new modules but that's something I could figure out if we want to know before merging.

I believe those are most of the concerns that I had. I haven't had any issues resolving dependencies though it is worth trying from a clean maven repository. There are also a few dependencies which use a different version from their JAR in the repository, though I haven't had issues yet with the deployment.

No

@mikejritter tested locally using cspace-ui.js and with the integration tests


axb21:
On first glance this looks OK to me. I'd like to play with it in my environment a bit before saying with confidence it looks good to merge. I'll work on that today/tomorrow.

axb21:
I built and deployed both `develop` and this branch, then compared the bundles and lib subdirectories. There are 57 jars in the `develop` deployment that are not present in `dryd-1754` (version number and .jar extension removed since there are jars that only differ by version). These seem to be the most likely source of potential problems:
- nuxeo-server/bundles/nuxeo-core-redis
- nuxeo-server/bundles/nuxeo-core-storage-mongodb
- nuxeo-server/lib/ant
- nuxeo-server/lib/ant-launcher
- nuxeo-server/lib/apacheds-bootstrap-extract
- nuxeo-server/lib/apacheds-btree-base
- nuxeo-server/lib/apacheds-constants
- nuxeo-server/lib/apacheds-core
- nuxeo-server/lib/apacheds-core-shared
- nuxeo-server/lib/apacheds-jdbm-store
- nuxeo-server/lib/apacheds-kerberos-shared
- nuxeo-server/lib/apacheds-protocol-shared
- nuxeo-server/lib/apacheds-schema-bootstrap
- nuxeo-server/lib/apacheds-schema-registries
- nuxeo-server/lib/apacheds-utils
- nuxeo-server/lib/arq
- nuxeo-server/lib/batik-all
- nuxeo-server/lib/commons-digester
- nuxeo-server/lib/commons-jexl
- nuxeo-server/lib/commons-pool2
- nuxeo-server/lib/commons-validator
- nuxeo-server/lib/flute
- nuxeo-server/lib/iri
- nuxeo-server/lib/jackson-annotations
- nuxeo-server/lib/jackson-databind
- nuxeo-server/lib/jackson-jaxrs
- nuxeo-server/lib/jackson-xc
- nuxeo-server/lib/jaudiotagger
- nuxeo-server/lib/javax.json
- nuxeo-server/lib/jbig2-imageio
- nuxeo-server/lib/jboss-seam
- nuxeo-server/lib/jcip-annotations
- nuxeo-server/lib/jdbm
- nuxeo-server/lib/jedis
- nuxeo-server/lib/jena
- nuxeo-server/lib/jersey-json
- nuxeo-server/lib/jettison
- nuxeo-server/lib/jolokia-core
- nuxeo-server/lib/jopt-simple
- nuxeo-server/lib/json-simple
- nuxeo-server/lib/jsr250-api
- nuxeo-server/lib/jstl
- nuxeo-server/lib/mina-core
- nuxeo-server/lib/nuxeo-core-storage-sql-extensions
- nuxeo-server/lib/nuxeo-stream
- nuxeo-server/lib/plexus-utils
- nuxeo-server/lib/shared-asn1
- nuxeo-server/lib/shared-ldap
- nuxeo-server/lib/shared-ldap-constants
- nuxeo-server/lib/stax-api
- nuxeo-server/lib/taglibs-standard-impl
- nuxeo-server/lib/tomcat-jdbc
- nuxeo-server/lib/tomcat-juli
- nuxeo-server/lib/vaadin-sass-compiler
- nuxeo-server/lib/webdav-jaxrs
- nuxeo-server/lib/wro4j-core
- nuxeo-server/lib/xml-apis-ext

I know it'll be a bit time-consuming but I think we should probably go through these to figure out what they do and ensure we really don't need them. The worry is that we'll hit weird runtime errors during uncommon workflows because the necessary runtime dependencies are not available. I poked at a few:
- `apacheds-core` is a directory server that includes LDAP, but has other things. Is it possible our security layer uses this?
- `jaudiotagger` is an audio file tagging library. Does media handling use this?
- `jolokia-core` is a debugging library that lets you attach to JMX over HTTP.
- `wroa4j-core` is "Web Resource Optimizer for Java". It doesn't look like we use it after a first glance but I'm not 100% sure

Another thing we might do is poke through the `git` history to see if we can find where each of these was added to the repository. Sometimes the commit message will give a clue as to why it was added. It's not always that helpful:
```bash
$ git log --follow --diff-filter=A --find-renames=40% -- ./3rdparty/nuxeo/nuxeo-server/9.10-HF30/lib/wro4j-core-1.7.7.jar
commit 6a257099d46284dac4bb23990a3c421e753d3fcd
Author: Ray Lee <ray.lee@lyrasis.org>
Date:   Wed Apr 17 18:31:13 2019 -0700

    DRYD-630: Upgrade nuxeo to 9.10-HF30.
```

mikejritter:
@axb21 I have notes on most of the dependencies which were removed though they aren't as granular as I would have liked (it was mostly for testing startup/api responses with the dependencies removed + some scattered notes). I'm not sure if doing it here would be the best, or if it would be easier to have the discussion somewhere else since there's so much to go over. Regardless I think I'll get my notes in a more readable format and show the dependency trail so we can see what nuxeo modules were bringing in the dependencies.

I have two jira tickets in my notes about the dependencies - DRYD-669 and DRYD-630. There isn't a ton of information in them but DRYD-669 did point to the batik dependency missing, though I couldn't reproduce it. I'll probably include that as I write up what I have so far.

axb21:
> I have notes on most of the dependencies which were removed though they aren't as granular as I would have liked (it was mostly for testing startup/api responses with the dependencies removed + some scattered notes). I'm not sure if doing it here would be the best, or if it would be easier to have the discussion somewhere else since there's so much to go over. Regardless I think I'll get my notes in a more readable format and show the dependency trail so we can see what nuxeo modules were bringing in the dependencies.

Sounds good. I feel like it'd make sense to create a spreadsheet and we can split the work of tracking down what the dependency does and ticking off whether we've checked that we don't need it, or something along those lines.

mikejritter:
We've gone through and categorized all the transitive dependencies which are now removed from the `lib` directory, and we now feel comfortable testing this in a larger environment so that SSO, reports, batch jobs, etc can all be tested.

PR #463, opened 2025-05-02
DRYD-1747: Add update_userid function to database

axb21:
**What does this do?**
This PR adds:
- A SQL script `services/src/main/resources/db/postgresql/update_userid_function.sql`
- A new `ant` target `create_update_userid_db_function` that calls this SQL script to create the `update_userid` function in the database

This PR updates:
- `ant deploy` to put `update_userid_function_sql` in  `$CSPACE_JEESERVER_HOME/cspace/services/db/postgresql` 
- `ant create_db` to run this script and create the `update_userid` function.

On initial `ant create_db`, the function will be added to the database. If one only wants to create the function and not call any other ant targets, `ant create_update_userid_db_function` will do that. One could also run the SQL script.

Jira ticket: https://collectionspace.atlassian.net/browse/DRYD-1747 . This is a request from hosting.

- Ensure there is no file `update_userid_function_sql` in `$CSPACE_JEESERVER_HOME/cspace/services/db/postgresql`. Run `ant deploy` in `services` and verify `update_userid_function_sql` is now present in `$CSPACE_JEESERVER_HOME/cspace/services/db/postgresql`
- Connect to the CollectionSpace postgress database In `psql`, run `\df`, and observe the `update_userid` function is not defined; run `DROP FUNCTION update_userid;` if it is. Issue `ant create_update_userid_db_function` in `services` and run `\df` again to verify `update_userid` is now defined 
- After a fresh `ant create_db`, connect to the CollectionSpace postgres database in `psql` and issue `\df`; observe the `update_userid` function is defined
- Run the `update_userid` function in `psql` (`SELECT update_userid('FOO','bar');` where `FOO` is a username) and verify the username is updated in the `accounts_common.userid`, `accounts_roles.user_id`, and `users.username` columns (CAUTION: this is a destructive and irreversible operation)

The `ant` build targets will overwrite a function named `update_userid` if it already exists in the database.

No.

@axb21 tested this locally.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -309,7 +309,7 @@
 	<target name="setup_initdb.sql" description="copy init_db scripts, replacing keywords">
 		<mkdir dir="${jee.server.cspace}/cspace/services/db/postgresql"/>
 		<copy todir="${jee.server.cspace}/cspace/services/db/postgresql">
-			<fileset dir="${jee.server.cspace}/cspace/services/db/postgresql" includes="drop_cspace_db.sql, init_cspace_db.sql, drop_nuxeo_db.sql">
+			<fileset dir="${jee.server.cspace}/cspace/services/db/postgresql" includes="drop_cspace_db.sql, init_cspace_db.sql, drop_nuxeo_db.sql, update_userid_function.sql">
 				<present targetdir="${src}/main/resources/db/postgresql" />
 			</fileset>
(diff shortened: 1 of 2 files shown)


mikejritter (approved):
Tested this locally and was able to see the in my cspace_default db. Didn't check the function but it looks simple enough and was given to us from hosting so I trust that it works.

PR #473, opened 2025-07-17
DRYD-1829: Remove flickr urls from tests

mikejritter:
**What does this do?**
This starts to add an integration test setup using the `maven-failsafe-plugin` in addition to the `jetty-maven-plugin`, and `maven-remote-resources-plugin`. It adds a place for the images hosted on flickr to live and includes some utilities to help with testing. 

The changes are:
* Add maven-failsafe-plugin
* Add jetty-maven-plugin
  * Configure jetty plugin to serve static files
* Add maven-remote-resources-plugin
* Create StaticImage enum for tests
* Rename *Test classes to *IT
* Remove flickr from Blob/Media tests

Jira: https://collectionspace.atlassian.net/browse/DRYD-1829

This updates the tests which require external urls to use an embedded jetty server, removing our dependency of flickr for testing. I opted not to put too much configuration in the top level pom even though it adds a bit of redundant config for some of the plugins.

In order to store the images, a new module `blob-test-utils` is added which includes an enum to help with creating urls for the images. The `maven-remote-resources-plugin` is used to bundle the images under `src/main/resources/` into its jar. Initially I wanted to use `src/test/resources`, but I'm not sure if this makes a difference.

The `blob`, `media`, and `restrictedmedia` modules all make use of the `blob-test-utils` and have been updated to use jetty to serve the images. Tests were renamed from `*Test.java` to `*IT.java` to make exclusion/inclusion as easy as possible in the `maven-surefire-plugin` configuration.

* Do a clean install collectionspace with the following tenants enabled
  * anthro, botgarden, core, materials, testsci
* If you have the `CSPACE_CORE_UI_BASE_URL_OPT` environment variable set, unset it as it will cause the json integration tests to fail
* Start collectionspace and wait the eternally long startup process 
* Run `mvn verify` and ensure all tests pass

The cspace-nightly-build will need to be updated to switch from `mvn test` to `mvn verify` in order to run all tests

Documentation needs to be updated to switch from `mvn test` to `mvn verify`

@mikejritter ran locally, but forgot to unset the `CSPACE_CORE_UI_BASE_URL_OPT` environment variable


services/blob/client/src/test/java/org/collectionspace/services/client/test/BlobServer.java
```diff
@@ -0,0 +1,79 @@
...
+
+import org.apache.http.entity.ContentType;
+import org.eclipse.jetty.server.Server;
+import org.eclipse.jetty.servlet.ServletHandler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class BlobServer {
+
+    private static final Logger logger = LoggerFactory.getLogger(BlobServer.class);
```
axb21: This probably shouldn't be static, because any failure to initialize this logged will result in catastrophic failure of this class to initialize, which in turn could lead to a catastrophic failure of anything depending on it (including the entire test suite potentially). Failure to initialize logging should not be a critical error, but a managed exception (or even ignored).

services/blob/client/src/test/java/org/collectionspace/services/client/test/BlobServer.java
```diff
@@ -0,0 +1,79 @@
...
+
+    private Server server;
+    private int port;
+
+    public void start() throws Exception {
+        ServerSocket
+        port = Integer.parseInt(System.getProperty("blob.test.port"));
```
axb21: Should put some failure management around this; if the `Integer.parseInt` fails here then `start()` will fail. A default test port could be assigned if that property isn't set.

axb21:
@mikejritter I reviewed the code but haven't tried running this locally yet. I'll do that after our meeting today and add comments if I notice anything.

services/blob/client/src/test/java/org/collectionspace/services/client/test/BlobServer.java
mikejritter: @axb21 This class was supposed to be removed, sorry about that, my rebase must have undone my change.

PR #474, opened 2025-08-26
Prototype advanced search endpoint

axb21:
**What does this do?**
This code introduces a new API endpoint to CollectionSpace services, `/advancedsearch`, which is a prototype of the in-progress advanced search feature. As currently implemented this endpoint accepts the same query parameters as the existing search, but returns a list of objects with a new set of fields (including the blob CSID of the object's thumbnail).

Addresses Jira ticket [Search Results > Create prototype backend service](https://collectionspace.atlassian.net/browse/DRYD-1826)

The API/endpoint should be tested with the concurrently developed front end.

None that I'm aware of. A new XSD schema is introduced, so `generate-sources` must be run during the build, but that is automatic in the default build process.

No.

@axb21 has run this code locally.


mikejritter (approved):
@axb21 Thanks for the work to get this started. I'm running it locally and everything seems to be in order.

PR #478, opened 2025-09-16
DRYD-1902: Update Strategy for Search Query

mikejritter:
**What does this do?**
The primary goal of this PR is to avoid querying the CollectionObjectResource for each csid and instead update CSpace in a way that allows us to get back the full CollectionObject XML. I'll fill this out with more information tomorrow. There's also some more refactoring still to do, but for the most part this is ready.

[Quick blurb about why the code is needed and a JIRA link. Do these changes meet the business requirements of the story?]

[Description of any tests that need to be performed once merged.]

[Description of any watchouts, dependencies, or issues we should be aware of.]

### DIFF
--- a/services/advancedsearch/service/src/main/java/org/collectionspace/services/advancedsearch/AdvancedSearch.java
+++ b/services/advancedsearch/service/src/main/java/org/collectionspace/services/advancedsearch/AdvancedSearch.java
@@ -12,6 +12,9 @@ import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.core.Request;
 import javax.ws.rs.core.UriInfo;
+import javax.xml.bind.JAXBContext;
+import javax.xml.bind.JAXBException;
+import javax.xml.bind.Unmarshaller;
 
 import org.collectionspace.collectionspace_core.CollectionSpaceCore;
 import org.collectionspace.services.advancedsearch.AdvancedsearchCommonList.AdvancedsearchListItem;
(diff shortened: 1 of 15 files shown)


spirosdi (approved):
just some minor notes, so I approve

services/advancedsearch/service/src/main/java/org/collectionspace/services/advancedsearch/AdvancedSearch.java
spirosdi, line 208: Not directly related to the changes, but we can use here Streams API to simplify the list processing:

```suggestion
return associatedMedia.getListItem().stream()
				.map(this::extractBlobCsid)
				.filter(blobCsid -> blobCsid != null)
				.collect(Collectors.toList());
				
private String extractBlobCsid(ListItem item) {
    return item.getAny().stream()
        .filter(el -> "blobCsid".equals(el.getTagName()))
        .map(Element::getTextContent)
        .findFirst()
        .orElse(null);
}
```

spirosdi, line 209: Not directly related to the updates, but you can use Stream API here:

```suggestion
if (associatedMedia == null || associatedMedia.getListItem() == null) {
			return new ArrayList<>();
		}

		return associatedMedia.getListItem().stream()
				.filter(item -> item != null && item.getAny() != null)
				.flatMap(item -> item.getAny().stream())
				.filter(el -> "blobCsid".equals(el.getTagName()))
				.map(Element::getTextContent)
				.filter(blobCsid -> blobCsid != null && !blobCsid.isEmpty())
				.collect(Collectors.toList());
```

spirosdi, line 209: Not directly related to the updates, but you can use Stream API here:

```suggestion
if (associatedMedia == null || associatedMedia.getListItem() == null) {
			return new ArrayList<>();
		}

		return associatedMedia.getListItem().stream()
				.filter(item -> item != null && item.getAny() != null)
				.flatMap(item -> item.getAny().stream())
				.filter(el -> "blobCsid".equals(el.getTagName()))
				.map(Element::getTextContent)
				.filter(blobCsid -> blobCsid != null && !blobCsid.isEmpty())
				.collect(Collectors.toList());
```

PR #479, opened 2025-09-22
DRYD-1818: Implement image ordering functionality for public browser

spirosdi:
**What does this do?**
* Adds a new denormalized field at the Media documents `objectCsid` holding the related object's ids.
* Maps the `updatedAt`, `identificationNumber`, `title` fields as sortable. 
* Maps the `objectCsid` field as keyword so that we can filter medias by related objects.

In the end we should be able to do a `POST /_search` from public browser and get the related medias of an object sorted by the sort param. The body of the request should contain:
```
{
    "query": {
        "terms": {
            "collectionspace_denorm:objectCsid": [{object-id}]
        }
    },
    "sort": 
        {
            "{sort-param}": "desc"
        }
}
```

We need to be able to search for medias, filtered by related objects and sorted by `updatedAt`, `identificationNumber` or `title`. https://collectionspace.atlassian.net/browse/DRYD-1818

Please verify that `POST localhost:9200/nuxeo_default/_search` calls with the following payload:
```
{
    "query": {
        "terms": {
            "collectionspace_denorm:objectCsid": [{"object-id"}]
        }
    },
    "sort": {
        {"media_common:title" | "media_common:identificationNumber" | "collectionspace_core:updatedAt"}: "desc"
    },
    "size":100,
    "terminate_after":100
}
```
responds with medias that are related to the specific `object-id` and sorted by the `sort` param passed.

No dependencies.

We will update the documentation when the relevant public-browser changes are also merged.

@spirosdi tested locally using local services and elastic search.

### DIFF
--- a/services/common/src/main/cspace/config/services/tenants/tenant-bindings-proto-unified.xml
+++ b/services/common/src/main/cspace/config/services/tenants/tenant-bindings-proto-unified.xml
@@ -1220,7 +1220,7 @@
 							"media_common:blobCsid",
 							"media_common:identificationNumber",
 							"media_common:title",
-							"media_common:altText",
+							"media_common:altText"
 						]
 					},
 					"properties" : {
--- a/services/common/src/main/cspace/config/services/tenants/anthro/anthro-tenant-bindings.delta.xml
+++ b/services/common/src/main/cspace/config/services/tenants/anthro/anthro-tenant-bindings.delta.xml
@@ -90,7 +90,10 @@
 							"ecm:currentLifeCycleState",
 							"ecm:name",
 							"ecm:primaryType",
-							"media_common:blobCsid"
+							"media_common:blobCsid",
+							"media_common:identificationNumber",
+							"media_common:title",
+							"media_common:altText"
 						]
 					},
 					"properties" : {
@@ -111,6 +114,10 @@
 						},
 
 						"collectionspace_core:createdAt": {
+							"type": "date",
+							"format": "date_time"
+						},
+						"collectionspace_core:updatedAt": {
 							"type": "date",
 							"format": "date_time"
 						},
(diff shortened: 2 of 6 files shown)


services/common/src/main/cspace/config/services/tenants/tenant-bindings-proto-unified.xml
```diff
@@ -1217,7 +1217,10 @@
 							"ecm:currentLifeCycleState",
 							"ecm:name",
 							"ecm:primaryType",
-							"media_common:blobCsid"
+							"media_common:blobCsid",
+							"media_common:identificationNumber",
+							"media_common:title",
+							"media_common:altText",
```
mikejritter, line 1223: Dangling `,` here

mikejritter, line 1223: We'll need to add these (and the other changes to the bindings) to each tenant binding which overrides the `tenant:elasticSearchIndexConfig` section. Unfortunately when the bindings are created for each tenant most sections just use `merge:action="replace"` which as you might guess will replace the entire section. If you want to verify you can check the merged configuration in `${tomcat}/cspace/config/services/tenants/${tenant}/tenant-bindings.merged.xml`.

It looks like the following files override the config:
* services/common/src/main/cspace/config/services/tenants/anthro/anthro-tenant-bindings.delta.xml
* services/common/src/main/cspace/config/services/tenants/bonsai/bonsai-tenant-bindings.delta.xml
* services/common/src/main/cspace/config/services/tenants/fcart/fcart-tenant-bindings.delta.xml
* services/common/src/main/cspace/config/services/tenants/lhmc/lhmc-tenant-bindings.delta.xml
* services/common/src/main/cspace/config/services/tenants/materials/materials-tenant-bindings.delta.xml

mikejritter (changes requested):
Overall this works as expected. I first tried to reindex my elasticsearch by POSTing to `http://localhost:8180/cspace-services/index/elasticsearch` which is how I noticed the dangling `,` in the index. Once that was fixed I was able to see the changes and verify they worked.

PR #516, opened 2026-03-24
Start Upgrade Process

mikejritter:
**What does this do?**
* This folds in the work done last year which started to look at changes needed for upgrading

No Jira.

This just brings the starting point of the upgrade work to be where it was last year. 

I rolled back a few changes to make this a little easier to manage:
* Formatting changes to poms are mostly ignored. Formatting in general is all over the place so it ends up creating more noise than anything.
* Reverted the nuxeo version change. That can be handled with the nuxeo updates, and old dependencies can continue to be resolved in the mean time.

This is not intended to be functional yet. Doing a `mvn dependency:resolve` should be sufficient, though there are still many changes to make so it won't really give us much information. 

As mentioned above, this is not close to being ready for production.

Eventually we'll need to update the installation/build instructions.

@mikejritter has been working locally with it

n/a


mikejritter:
@spirosdi if you want to look this over you can. It's getting things ready more than anything, and what I've been working on top of. I'm not sure of all the changes it really had anymore but I believe it was mostly a few dependency changes and plugin updates.

PR #517, opened 2026-03-26
DRYD-2052: Update jaxb plugin and schemas

mikejritter:
**What does this do?**
* Updates our XSDs to use the jakarta namespace
* Update version and configuration for the jaxb-maven-plugin
* Add dependencies for jaxb modules

Jira: https://collectionspace.atlassian.net/browse/DRYD-2052

This is an update to the classes which are created through the jaxb-maven-plugin. It allows them to use jakarta imports.

I updated the dependencies in the jaxb modules as appropriate to try and remove any redundant/unnecessary dependencies. Some which required `AbstractCommonList` will still depend on `org.collectionspace.services.jaxb`, but others only specify `jakarta.xml.bind-api` and `jaxb-plugins-runtime`. Likewise some of the authority jaxb modules still require `org.collectionspace.services.authority.jaxb`.

This is a bit tricky because the project still does not compile. A few things can be done to spot check the changes:
* Verify that all XSDs in the repository no longer reference the java.sun.com in the namespace
  * Running `grep -lr --include '*.xsd' 'http://java.sun.com'` at the root of the repository will show all files which still reference it.
  * This set should only include the modules which use hyperjaxb for JPA entity generation: account, authentication, authorization, and hyperjaxb
  * **Note:** The `note` module is also included in here, but the module is not used in compilation at all. We should probably remove it.
* Verify that `mvn generate-sources` completes successfully
  * In addition, generated classes can be seen with `find . -wholename '*/target/generated-sources/*'`
  * As I've been building frequently, I'm not sure if the modules with a dependency on `org.collectionspace.services.jaxb` module will complete. In that case, a `mvn install` might need to be done in that module.

Many changes still to be done for production :)

No

n/a


mikejritter:
@spirosdi if this is too big I'm sure we can find a way to split it up to be more manageable; most of the changes were just find/replace though so no big logic. Maybe something like plugin configuration -> base jaxb module -> collectionobject + procedures -> authorities.

PR #520, opened 2026-04-06
Migrate account from xsd to classes

mikejritter:
**What does this do?**
* Removes hyperjaxb plugin from the account jaxb module
* Create classes for what were previously schema generated
* Add jspecify for some null safety

Jira: https://collectionspace.atlassian.net/browse/DRYD-2074

This migrates both the JPA entities and JAXB models as we are not migrating to jakarta persistence yet, so using the plugin for code gen is no longer available afaik (you can do either all javax or all jakarta).

JSpecify has been added in order to start to help understand what can/cannot be null. At the moment, I only added NonNull to the List getters in order to reflect previous behavior.

* On dev: 
  * Compile the account/jaxb module
  * copy the generated sources to a separate directory (e.g. `diff`)
* switch to this branch
  * compare the old generated classes to the new classes, e.g. `git diff --no-index <(cat diff/org/collectionspace/**/*.java) <(cat src/main/java/org/collectionspace/**/*.java)`
  * `mvn clean compile` should pass (might have dependencies which need to be compiled as well)

Still more modules to migrate + other javax to jakarta import migration before compilation will be possible.

There's also an old Aspect which we should look at once the migrations are finished (JaxbXMLGregorianCalendarMarshal).

n/a


mikejritter:
Found one issue where some fields were being directly accessed. Pushed a commit to use their getters.

PR #526, opened 2026-04-16
DRYD-2092: Make SecurityInterceptor Implement ContainerRequestFilter

mikejritter:
**What does this do?**
* Migrate SecurityInterceptor to Container{Request,Response}Filter 
* Remove parameters from allowAnonymousAccess
* Misc cleanup, e.g. parameterized logging, pointless booleans, etc

Jira: https://collectionspace.atlassian.net/browse/DRYD-2092

This is required for the upgrade process, but is something which is possible without changing any dependencies so it makes sense to do it separate and get some extra testing. Fairly straightforward change going from PreProcessInterceptor to ContainerRequestFilter and PostProcess to Response.

* Rebuild collectionspace
* Verify that any API call which requires a nuxeo login still works, e.g. advancedsearch

I actually noticed the behavior for this interceptor is a little... odd. As in it appears to be running multiple times per request (I'm not sure if it's after every filter or something else). I verified this is not a result of the changes and I believe the unbalanced login message points to this being a thing for some time. Perhaps during the RESTEasy update I'll have a chance to look at it further.

n/a

### DIFF
--- a/services/common/src/main/java/org/collectionspace/services/common/security/SecurityInterceptor.java
+++ b/services/common/src/main/java/org/collectionspace/services/common/security/SecurityInterceptor.java
@@ -32,10 +32,6 @@ import java.lang.reflect.InvocationTargetException;
 import java.security.Principal;
 import java.util.HashMap;
 import java.util.Set;
-
-import org.jboss.resteasy.spi.Failure;
-import org.nuxeo.runtime.api.Framework;
-
 import javax.security.auth.Subject;
 import javax.security.auth.login.LoginContext;
 import javax.security.auth.login.LoginException;
@@ -43,6 +39,8 @@ import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.container.ContainerRequestFilter;
 import javax.ws.rs.container.ContainerResponseContext;
 import javax.ws.rs.container.ContainerResponseFilter;
+import javax.ws.rs.container.ResourceInfo;
+import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Request;
 import javax.ws.rs.core.Response;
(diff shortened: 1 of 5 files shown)


services/common/src/main/java/org/collectionspace/services/common/security/SecurityInterceptor.java
```diff
@@ -64,17 +61,17 @@
...
+import org.nuxeo.runtime.api.Framework;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
  * RESTeasy interceptor for access control
  * @version $Revision: 1 $
  */
-@SecurityPrecedence
-@ServerInterceptor
```
spirosdi, line 75: Can you somehow confirm that removing the `SecurityPrecedence` annotation doesn't cause any priority race condition issue?

spirosdi:
Confirmed that unauthorised `POST /cspace-services/collectionobjects` or `GET /cspace-services/workauthorities` give `401` as expected.

spirosdi (approved):

services/common/src/main/java/org/collectionspace/services/common/security/SecurityInterceptor.java
mikejritter: @spirosdi looking at both `develop` and this branch, at runtime I see the same execution stack for the filters: `CSpaceUserAttributeFilter` -> `CSpaceFilter` -> `JsonToXmlFilter` -> `XmlToJsonFilter` -> `SecurityInterceptor`.

One thing we could do to ensure ordering, especially for the future, is to add the `@Priority` annotation with the `AUTHENTICATION` priority, which I believe is equivalent to what the previous annotation was doing.

The `CSpaceFilter` also looks like a good candidate for removal.

mikejritter: I went and added the `Priority` annotation, as it is a small change that seems worth it.

PR #527, opened 2026-04-16
DRYD-2093: Replace HttpResponseCodes with JAX-RS Equivalent

mikejritter:
**What does this do?**
* Replaces HttpResponseCodes with jax-rs equivalent

Jira: https://collectionspace.atlassian.net/browse/DRYD-2093

Not entirely necessary for the upgrade iirc but makes things a little bit tidier.

These changes are fairly minor so shouldn't require too much. Maybe just validating some of the request methods that were changed still respond correctly (integration tests should cover this).

* Create an account and role
* Delete the role
* Delete the account

There might be some dependencies we can remove, and some changes I had made in my other branch snuck in b/c this was cherry picked. Otherwise none. 

n/a

@mikejritter recompiled

n/a


services/common/src/main/java/org/collectionspace/services/common/relation/RelationResource.java
```diff
@@ -146,26 +142,26 @@ private RelationsCommonList getRelationList(
...
-	    	
-	    	if (csidList.isEmpty() == false) {
+
+	    	if (!csidList.isEmpty()) {
 	            ServiceContext<PoxPayloadIn, PoxPayloadOut> ctx = createServiceContext();
 	        	DocumentHandler<PoxPayloadIn, PoxPayloadOut, DocumentModel, DocumentModelList> handler = createDocumentHandler(ctx);
 	            getRepositoryClient(ctx).delete(ctx, csidList, handler);
 	    	} else {
-	            result = Response.status(HttpResponseCodes.SC_NOT_FOUND).build();
+	            result = Response.ok().build();
```
spirosdi: With this change a DELETE that finds nothing to delete now returns 200 instead of 404. Please change to a NOT_FOUND response.

services/common/src/main/java/org/collectionspace/services/common/NuxeoBasedResource.java
```diff
@@ -308,7 +307,7 @@ protected Response delete(ServiceContext<PoxPayloadIn, PoxPayloadOut> ctx, Strin
             throws Exception {
     	DocumentHandler<PoxPayloadIn, PoxPayloadOut, DocumentModel, DocumentModelList> handler = createDocumentHandler(ctx);
         getRepositoryClient(ctx).delete(ctx, csid, handler);
-        return Response.status(HttpResponseCodes.SC_OK).build();
+        return Response.status(Response.Status.OK).build();
```
spirosdi: maybe this needs to be changed to `Response.ok().build()` to be consistent with the rest of the changes.

spirosdi (changes requested):
A couple of minor but necessary changes

spirosdi (approved):

PR #528, opened 2026-04-20
DRYD-2094: Replace ClientResponse with Response

mikejritter:
**What does this do?**
* Replace ClientResponse with javax Response
* Remove unused ClientResponse imports
* Update antrun plugin config 

Jira: https://collectionspace.atlassian.net/browse/DRYD-2094

This is another small change for RESTEasy 3.x before we upgrade so that we can deploy to dev and do some early testing there. This should also make the move to the jakarta namespace a bit easier as the imports are all the same.

When running the integration tests there was an error in the AccountClientTest related to the hibernate.cfg.xml and caused the update to the plugin in the pom.xml.

Most of these have no functional changes that need to be tested. Only the ClaimProxy had return types updated should have its test suite ran to check as that is the only usage.

For the plugin configuration, in the project directory for `services/account/client`, run `mvn clean package -DskipTests` and see that the hibernate.xml.cfg which is written to `target/test-classes` has correct settings for the database configuration. The tests in `services/account/client` can be ran as well.

None

No

@mikejritter ran the integration tests locally

n/a

### DIFF
--- a/services/account/client/pom.xml
+++ b/services/account/client/pom.xml
@@ -76,34 +76,26 @@
         <finalName>cspace-services-account-client</finalName>
         <plugins>
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-antrun-plugin</artifactId>
+                <version>3.1.0</version>
                 <executions>
                     <execution>
                         <id>subs-hibernate-config</id>
-                        <phase>generate-test-resources</phase>
+                        <phase>process-test-resources</phase>
                         <goals>
                             <goal>run</goal>
                         </goals>
                         <configuration>
-                            <tasks>
+                            <target>
                                 <property name="runtime-classpath" refid="maven.runtime.classpath"/>
                                 <ant target="setup_hibernate.cfg" inheritRefs="true"/>
-                            </tasks>
+                            </target>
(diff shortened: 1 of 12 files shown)


spirosdi (approved):
Confirmed that the plugin configuration change behaves as described. Does it make sense to apply the change also to the other modules where the pattern (`<tasks>` + generate-test-resources) exists, that also run setup_hibernate.cfg? Like `services/authentication/client/pom.xml, services/account/pstore/pom.xml, services/authorization/pstore/pom.xml, services/authentication/pstore/pom.xml, services/security/client/pom.xml`? Even though only the `services/security/client` actually has tests. And they were green on my side. Not necessary to resolve in this PR (maybe just in a follow-up task).

`services/note/client/...NoteProxy.java` still has `ClientResponse<NotesCommonList>`. Is this intentional?

All minor notes, so I approve

mikejritter:
@spirosdi re: the NoteProxy, this is something I've been meaning to bring up. The note modules are commented out in the `services/pom.xml` file with very little reasoning as to why from the old jira associated w/ the commit. IMO it's probably worth just removing the modules since they don't seem to be in use.

I'll create the additional task for the other modules to bundle in with the other upgrade work.

PR #532, opened 2026-05-05
DRYD-2101: Update Authority Fulltext Queries

mikejritter:
**What does this do?**
* Update fulltext query for authority refs
* Add boolean parameter for toggling cleaning of keywords
* Various cleanup (type inf, modifier sorting, etc)

https://collectionspace.atlassian.net/browse/DRYD-2101

The way fulltext data is extracted has changed which has caused a few APIs to return incorrect results. This was being caused by assumptions about the structure of the fulltext column and what the data looks like. The update changes how the query is being made to only use the authority, authority instance name, and the short id of the authority. This allows us to still identify specific instances of authorities (e.g. `vocabularies inventorystatus instorage`. 

In addition, by changing from a phrase search to a normal fulltext search, our own query builder was modifying the keywords and changing the structure of the query. This would have an impact on terms like `not_approved` which would become `-approved`. The `clean` parameter was added so that we skip the preprocessing which strips punctuation and searches for additional keywords. It might be worth reviewing our keyword searching to see if we want to make changes to what is allowed again.

* Create some procedures and authorities and relate them
* In the authorities, see that the `Terms Used` and `Used By` panels populate appropriately
* When searching, verify that the keyword search returns the expected results

Might be some follow up tasks for this. Also need to check if the old indexing will continue to work with how the queries have changed.

No. We should check the current keyword search docs to see if anything needs to change.

n/a

### DIFF
--- a/services/client/src/main/java/org/collectionspace/services/client/IQueryManager.java
+++ b/services/client/src/main/java/org/collectionspace/services/client/IQueryManager.java
@@ -30,93 +30,91 @@ import java.util.List;
 
 public interface IQueryManager {
 
-	final static String SEARCH_COMBINE_QUERY_PARAM = "combine";
-	final static String SEARCH_COMBINE_AND = "and";
-	final static String SEARCH_COMBINE_OR = "or";
-	final static String SEARCH_GROUP_OPEN = "(";
-	final static String SEARCH_GROUP_CLOSE = ")";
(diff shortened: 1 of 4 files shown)


mikejritter:
I've been working on the Integration Tests in parallel to this and didn't get any issues in them so I'm going to merge this.

PR #533, opened 2026-05-07
DRYD-2102: Fixes for Thumbnail Derivative Generation

mikejritter:
**What does this do?**
* Replace deprecated FileManager APIs
* Handle view computation for images
* Add configuration for the Nuxeo Stream component
* Add configuration disable Nuxeo's PictureChangedListener
* Misc cleanup

Jira: https://collectionspace.atlassian.net/browse/DRYD-2102

When upgrading Nuxeo, changes were made to how their derivative generation works. This changes our workflow to stop relying on the Nuxeo EventListener which add the views/derivatives and instead does the work immediately after creating the Document for the Blob.

logmanager-config.xml adds configuration for the Nuxeo Stream log. This currently uses the in memory implementation to avoid needing to deploy Kafka.

* Rebuild and start CollectionSpace
* Upload a new MediaHandling record with an attached blob
* Refresh the page and see that the derivatives were generated 
  * Note: disabling the cache might be necessary so that the server is queried for the updated images.

None.

No.

@mikejritter tested locally. It might be good to double check the blob integration tests as well to verify the generated images are correct.


services/common/src/main/java/org/collectionspace/services/common/imaging/nuxeo/NuxeoBlobUtils.java
```diff
@@ -683,16 +690,53 @@ private static DocumentModel createDocumentFromBlob(
...
-
+			final var context = FileImporterContext.builder(repoSession.getCoreSession(), inputStreamBlob, blobLocation)
+												   .overwrite(overwrite)
+												   .fileName(blobName)
+												   .build();
 			FileImporter defaultFileImporter = fileManagerService.getPluginByName("DefaultFileImporter");
-			result = defaultFileImporter.create(
-					repoSession.getCoreSession(), inputStreamBlob, blobLocation, overwrite, blobName, getTypeService());
+			result = defaultFileImporter.createOrUpdate(context);
+		}
```
mikejritter: I'm pretty skeptical of if this is really going to do anything other than updating the inputStreamBlob. If the `createOrUpdate` above fails I can't imagine this would suddenly pass. Otherwise this might be run when called from the PublicItemUtil, which as far as I can tell doesn't seem to be used by the ui anymore.

I think it might make sense to just do a check to update the `inputStreamBlob` and `computeViews`, and merge the creation of the `result` Document.

spirosdi: Sure, that makes sense.

spirosdi:
I tested it locally and the derivatives are generated fine. 

Initially I tested it with a damaged jpg: 
image_7
the `POST http://localhost:8180/cspace-services/blobs` gave a 400: `Create request failed: No payload found for 'CREATE' action.`. Uploading the same file in core.dev worked. But since it is a damaged file in any case, it doesn't really matter. Just in case it helps you debugging or spotting an edge case.

PR #536, opened 2026-05-14
Update ES Mappings and Analysis config

mikejritter:
**What does this do?**
* Remove _all meta field
* Change nGram token filter to ngram

Part of https://collectionspace.atlassian.net/browse/DRYD-2111 but not a full test. I wanted to make sure that our deployment will start when ElasticSearch is enabled.

Relevant ES docs:
* _all meta removal - https://www.elastic.co/guide/en/elasticsearch/reference/7.0/breaking-changes-7.0.html#all-meta-field-removed
* nGram token filter - https://www.elastic.co/guide/en/elasticsearch/reference/7.0/breaking-changes-7.0.html#deprecated-ngram-edgengram-token-filter-cannot-be-used

* Rebuild cspace to get the updated config
* Set `elasticsearch.enabled` to `true` before starting CollectionSpace
* See that the index is created and the ES component starts

We should go through both the mapping and analysis settings again to see if there are any other changes we need to make.

If we have ES configuration we will need to update it.

@mikejritter tested using ES 7.17.29

n/a

### DIFF
--- a/3rdparty/nuxeo/nuxeo-server/9.10-HF30/config/proto-elasticsearch-extension.xml
+++ b/3rdparty/nuxeo/nuxeo-server/9.10-HF30/config/proto-elasticsearch-extension.xml
@@ -56,9 +56,9 @@
               "type": "path_hierarchy"
             },
             "ngram_tokenizer": {
-              "type": "nGram",
+              "type": "ngram",
               "min_gram": 3,
-              "max_gram": 12
+              "max_gram": 3
             }
           },
           "analyzer": {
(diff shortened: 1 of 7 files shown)


mikejritter:
@spirosdi I realize I forgot to tag you on this. These are the ES changes which were added to the upgrade branch to allow the server to start with `elasticsearch.enabled=true`. I thought I had seen something about the max_gram size as well which caused me to update ours but I'm no longer sure what I was looking at. It might be worth looking at our settings more as we test.

PR #537, opened 2026-05-18
DRYD-2114 CI/CD > JS Packages > Auto-resolve frontend patch versions

spirosdi:
**What does this do?**
It drops the patch version of cspace-ui.js, public browser and profile plugins inside the different build.properties files.

Doing so allows us to auto resolve the latest patch version during deployment, without the need to build and release a whole new cspace version: https://collectionspace.atlassian.net/browse/DRYD-2114

You don't really need to rebuild so that the changes have effect. You can just run the `deploy_tenants` ant target, which calls `download_cspace_ui_js` and `download_tenant_js` for each tenant. Please run it inside the cspace-ui directory:
`ant deploy_tenants`. Then you may check the welcome screens of the different tenants: http://localhost:8180/cspace/core/welcome, http://localhost:8180/cspace/anthro/welcome etc... And confirm that the "UI version" and "{Plugin} profile version" are set correctly. 

For public browser, please run the `deploy_cspace_public_browser_js` ant target and check in the http://localhost:8180/cspace/core/public/search page that the correct script source is being set.

I created the branch from `main` so that we are able to release the changes before the next release. But let us first merge it to `develop` and test also there.

The https://github.com/dts-hosting/data-toolkit/issues/255 is related but not a blocker.

The https://collectionspace.atlassian.net/wiki/spaces/CPD/pages/3701473281/Release+Checklist should be updated accordingly. A new page describing the Acceptance Criteria of a patch version should also be added.

@spirosdi ran it locally

no new security vulnerabilities.


spirosdi:
converting to draft, as I just realised I haven't changed the public browser configuration.

mikejritter (approved):

PR #539, opened 2026-06-02
DRYD-2090: Updates to Nuxeo Deployment

mikejritter:
**What does this do?**
* Move nuxeo config into 2023 folder
* Add SMTP config to fix startup warnings
* Remove login config to fix startup warnings
* Update build properties for new versions of Nuxeo and Tomcat

Jira: https://collectionspace.atlassian.net/browse/DRYD-2090

These are updates for our deployment so that it is more consistent with the version of Nuxeo which we've updated to. In addition some config files have been added to handle some warnings which were being logged when Nuxeo starts. It does not address any of the warnings that occur when shutting down Tomcat.

* Rebuild and deploy
* Check either Tomcat's `catalina.out` or `cspace-services.log` and see that the previous log entries are no longer found
  * For reference, the smtp config related error was about the JNDI Mail resource: `Name [Mail] is not bound in this Context.`
  * The login config related error was `Warning: target extension point 'domains' of 'org.nuxeo.runtime.LoginComponent' is unknown.`

No dependencies. I'll run the integration tests again as a sanity check.

Also for releasing to production - I kept these changes pretty minimal. There are some additional components we can configure for Nuxeo which involve: metrics, in memory config for the blob transient storage, and probably some other things. As I mentioned in the related Jira, I think it's better to disable what we don't need by using the configuration rather than trying to omit dependencies.

I'm going to start working on docs about how we configure Nuxeo and where to find additional information for making changes.

@mikejritter tested locally. 

n/a


build.properties
spirosdi: `nuxeo.templates` is undefined here. If nothing consumes `nuxeo.templates.dir` makes sense to remove it completely.

spirosdi:
I still see the `The login config related error` in the logs. Maybe because the `login-config.xml` is not actually removed but moved to 2023/config?

mikejritter:
@spirosdi Yep, the login-config was supposed to be deleted. I just removed it now along with the unused build property.

spirosdi (approved):
both warnings are gone now after redeploying.

PR #538, opened 2026-05-28
DRYD-2114 CI/CD > JS Packages > Auto-resolve frontend patch versions

spirosdi:
Re-targeting original PR #537 to main. See #537 for full description.

PR #540, opened 2026-06-03
No Jira: Disable DTD when Parsing XML

mikejritter:
**What does this do?**
This cherry picks commit 5ecf54a into main

No jira. This disables the Doctype declaration on XML preventing external entities from being loaded.

* Rebuild and start
  * Reminder: this is the current release which uses JDK 8
* Try to POST or PUT to the collectionspace API with XML that includes a doctype declaration, e.g.
```
<?xml version="1.0"?>
<!DOCTYPE document []>
<document>
  <ns2:collectionobjects_common xmlns:ns2="http://collectionspace.org/services/collectionobject">
    <objectNumber>2027.0.1</objectNumber>
  </ns2:collectionobjects_common>
</document>
```
* See that server rejects the request

This includes the version bump for 8.3.1 so maybe #538 should be merged first. We might also want to create a ticket for creating tests for this. Although the manual check is fine, something more comprehensive would probably be good.

No

@mikejritter tested locally

This handles the same vuln from #535

### DIFF
--- a/services/client/src/main/java/org/collectionspace/services/client/PoxPayload.java
+++ b/services/client/src/main/java/org/collectionspace/services/client/PoxPayload.java
@@ -4,6 +4,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.StringReader;
+import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


spirosdi (approved):
Looks good. Also tested locally and works as expected

spirosdi:
Let us release the 8.3.1 without the #538 . Those changes are not really necessary for lyrasis hosted clients. Their dependency resolution is handled in a completely different way (through the plugin version map file) and they are already updated. 

Let us release the #538 as a next patch release mainly for self-hosted clients and demo/qa sites. Actually for the self hosted ones we only need to tag and release in github. I think this is also valid for demo/qa environments as they actually download the ui dependencies from github. In any case I think it is safer to release the #538 separately as it is not that high prio and it is better not to delay the current PR updates because of any #538 implications.

PR #542, opened 2026-06-17
DRYD-2123   Nuxeo > Corrupted jars make build fail silently

spirosdi:
**What does this do?**
It replaces dead Nuxeo public/vendor release repos with cspace S3 mirror in nuxeo-server pom.

https://collectionspace.atlassian.net/browse/DRYD-2123

1. Please purge your local m2 cache
2. Clear the previously-deployed Nuxeo bundles/lib in Tomcat: `rm -rf "$CSPACE_JEESERVER_HOME/nuxeo-server/bundles"/* "$CSPACE_JEESERVER_HOME/nuxeo-server/lib"/*`
3. Rebuild and redeploy

Confirm that services start normal, the `java.util.zip.ZipException: error in opening zip file` runtime error should be gone.

No dependencies

Changelog has been updated

@spirosdi ran it locally

no new security vulnerabilities

### DIFF
--- a/3rdparty/nuxeo/nuxeo-server/pom.xml
+++ b/3rdparty/nuxeo/nuxeo-server/pom.xml
@@ -16,16 +16,16 @@
 
   <repositories>
     <repository>
-      <id>nuxeo-public-releases</id>
-      <url>https://maven-eu.nuxeo.org/nexus/content/repositories/public-releases</url>
+      <id>cspace-nuxeo-public-releases</id>
+      <url>https://cspace-maven-repo.s3-us-west-2.amazonaws.com/nuxeo/public-releases</url>
       <snapshots>
         <enabled>false</enabled>
       </snapshots>
     </repository>
 
     <repository>
-      <id>nuxeo-vendor-releases</id>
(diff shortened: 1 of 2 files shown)


3rdparty/nuxeo/nuxeo-server/pom.xml
```diff
@@ -16,16 +16,16 @@
 
   <repositories>
     <repository>
-      <id>nuxeo-public-releases</id>
-      <url>https://maven-eu.nuxeo.org/nexus/content/repositories/public-releases</url>
+      <id>cspace-nuxeo-public-releases</id>
+      <url>https://cspace-maven-repo.s3-us-west-2.amazonaws.com/nuxeo/public-releases</url>
```
mikejritter, line 20: @spirosdi in dev I had used the updated packges.nuxeo repository, but I wonder if we should update dev to point to our s3 repo as well. The packages would still need to be mirrored but it might save from any more future issues.

mikejritter (approved):
Tested by running deploy and checking the jars.

btw I noticed the plugin repository defined is _really_ slow. I think I removed in in dev, we could remove it here as well.

PR #547, opened 2026-07-01
DRYD-2124: Set Release to JDK 21

mikejritter:
**What does this do?**
Sets the release to target JDK 21

Jira: https://collectionspace.atlassian.net/browse/DRYD-2124

As part of the upgrade we wanted to see what version of the JDK we should land on. After research and some testing we decided on JDK 21.

* Rebuild the project with JDK 21 w/ the changes in https://github.com/collectionspace/application/pull/354
* Deploy CollectionSpace
* Use the application to some extent

None

The release/build docs need to be updated

@mikejritter is currently running the integration tests...

n/a

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,7 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<cspace.services.version>${revision}</cspace.services.version>
 		<cspace.services.client.version>${revision}</cspace.services.client.version>
-		<java.version>17</java.version>
+		<java.version>21</java.version>
 		<tomcat.version>9.0.99</tomcat.version>
 
 		<!-- CSpace managed dependencies -->


spirosdi:
The only issue I found is for MH records a `GET http://localhost:8180/cspace-services/blobs/cfac7b55-bbfa-4af5-a515/derivatives/Medium/content` gives a `400` with `Create request failed: java.lang.NullPointerException: Cannot invoke "org.collectionspace.services.common.blob.BlobOutput.getBlobsCommon()" because "blobOutput" is null`. This happens for MH records that existed before checking out the changes. For new MH records it works as expected.

spirosdi:
> Create request failed: java.lang.NullPointerException: Cannot invoke "org.collectionspace.services.common.blob.BlobOutput.getBlobsCommon()" because "blobOutput" is null

after checking the logs, I see that the issue is because of Tomcat switch and not fully copying Nuxeo store. So nothing to do with the JDK change.

spirosdi (approved):

PR #549, opened 2026-07-02
NOJIRA: Replace cspace-maven-repo

mikejritter:
**What does this do?**
Replaces cspace-maven-repo with cspace-program-production-cdn-cspace-maven

No jira. The s3 bucket with the maven repository is changing accounts and this is the new bucket name.

Rebuild cspace with this PR

None

No

I got a 403 when trying to build so I assume the change hasn't been made yet. Once it builds successfully I'll set this as being ready.

n/a

### DIFF
--- a/3rdparty/nuxeo/nuxeo-server/pom.xml
+++ b/3rdparty/nuxeo/nuxeo-server/pom.xml
@@ -17,7 +17,7 @@
   <repositories>
     <repository>
       <id>cspace-nuxeo-public-releases</id>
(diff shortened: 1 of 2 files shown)


mark-cooper:
@mikejritter the new bucket won't be (shouldn't be) publicly accessible, you'll want to go through the CloudFront url: https://cspace-maven.collectionspace.org (ask @mglyrasis for details).

mikejritter:
@mark-cooper Oh thanks, I was thinking about whether or not we could create a cname for it so we wouldn't have to deal with this in the future and was just looking at cloudfront for that. Would be nice if these things were communicated with us though.

spirosdi (approved):

PR #550, opened 2026-07-06
DRYD-2002: Password Complexity Requirements

mikejritter:
**What does this do?**
* Add tenant binding configuration for password requirements
* Add basic password requirement validators
* Add endpoint to get password requirement config
* Validate passwords during Account create and update requests
* Validate passwords during password reset requests
* Pass password requirement config to the service-ui template for form validation

Jira: https://collectionspace.atlassian.net/browse/DRYD-2002

This is for better compatibility for the HECVAT. It adds configurable password requirements through the tenant bindings and applies them to the Account create/update endpoints as well as the password reset request. For validation in the front end, the password requirement config is passed as a Map to the service-ui template which handles rendering of the form.

* Follow the steps in https://github.com/collectionspace/cspace-ui.js/pull/362

It should be noted this work is done on top of the #547 PR 

No. Documentation will need to be added.

@mikejritter tested locally

n/a

### DIFF
--- a/services/account/service/src/main/java/org/collectionspace/services/account/AccountResource.java
+++ b/services/account/service/src/main/java/org/collectionspace/services/account/AccountResource.java
@@ -92,6 +92,7 @@ import org.collectionspace.services.common.storage.TransactionContext;
 import org.collectionspace.services.common.storage.jpa.JpaStorageUtils;
 import org.collectionspace.services.config.ServiceConfig;
 import org.collectionspace.services.config.tenant.EmailConfig;
+import org.collectionspace.services.config.tenant.PasswordRequirementConfig;
 import org.collectionspace.services.config.tenant.TenantBindingType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -164,8 +165,9 @@ public class AccountResource extends SecurityResourceBase<AccountsCommon, Accoun
     }
 
     @GET
+    @Produces("application/xml")
     @Path(AccountClient.PASSWORD_COMPLEXITY_PATH)
-    public Response getPasswordComplexity(@Context UriInfo uriInfo) {
+    public PasswordRequirementConfig getPasswordComplexity(@Context UriInfo uriInfo) {
         ServiceContext<AccountsCommon, AccountsCommon> ctx;
         try {
             ctx = createServiceContext(null, AccountsCommon.class, uriInfo);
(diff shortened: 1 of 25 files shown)


services/account/service/src/main/java/org/collectionspace/services/account/AccountResource.java
```diff
@@ -284,6 +302,18 @@ public String processPasswordResetForm(@Context HttpServletRequest request) thro
         uiConfig.put("tenantId", tenantId);
         uiConfig.put("tenantLoginUrl", ConfigUtils.getUILoginSuccessUrl(tenantBinding));
 
+        PasswordRequirementConfig passwordRequirementConfig = tenantBinding.getPasswordRequirementConfig();
+        if (passwordRequirementConfig != null && passwordRequirementConfig.isEnabled()) {
+            uiConfig.put("passwordRequirements", Map.of(
+                "minLength", passwordRequirementConfig.getMinLength(),
```
spirosdi: `getMinLength()` is a nullable Integer (`minOccurs="0"` in tenant.xsd) so this throws a NPE when `minLength` is not set. I confirmed getting a `500` on the reset page.

spirosdi:
https://github.com/collectionspace/services/blob/b72b97bc766861fd3f0e8a1d4be68bee7a940081/services/common/src/main/java/org/collectionspace/services/common/security/SecurityUtils.java#L120-L133 should be skipped when tenant config is set. Otherwise min 8/max 24 runs on every create/update/reset and tenant config is ignored.

The legacy check error message includes the password:
image
We should definitely drop that.

spirosdi (changes requested):

services/account/service/src/main/java/org/collectionspace/services/account/validator/CSpacePasswordValidator.java
```diff
@@ -0,0 +1,57 @@
...
+        final var validators = initializeForTenant(binding);
+        return validators.stream()
+            .map(validator -> validator.validate(password))
+            .reduce(new ValidationResult(), ValidationResult::merge);
+    }
+
+    private static List<PasswordValidator> initializeForTenant(final TenantBindingType binding) {
+        final var config = binding.getPasswordRequirementConfig();
+
+        final var validators = new ArrayList<PasswordValidator>();
```
spirosdi, line 30: I propose that we add a `MaxBytesValidator` unconditionally here. To validate the max 72 bytes password length. The `BCryptPasswordEncoder` truncates the password to the first 72 bytes in any case. So instead of "failing" silently and misleading the user thinking that the whole password is being used, we can fail loudly when the password is longer than 72 bytes, with an `ERR_TOO_LONG` error status.

spirosdi (approved):
Single comment about a max-bytes unconditional validator, so I approve.

PR #552, opened 2026-07-15
NO-JIRA: Bugfix/reindex media on relate

spirosdi:
**What does this do?**
* Reindex the related Media when a Relation between Collection Object and a Media is created or deleted. So that the media's `collectionspace_denorm:objectCsid` field stays in sync with its relations.
* Handles both directions (subject=Media/object=CollectionObject and subject=CollectionObject/object=Media)

The public browser detail page image gallery finds media by querying `collectionspace_denorm:objectCsid` on Media docs. That field is only written when the Media doc itself is indexed, but relation create/delete events previously only reindexed the Collection Object. As a result newly related medias do not appear in the public browser until the media record is saved which triggers a reindex. Similarly, unrelated/deleted media keep showing until edited.

1. Create a media with an image, published to All
2. From the media record, relate it to a collection object that is published to the public browser. Without editing the media, verify the image appears on the object's public browser detail page.
3. Repeat the relation from the collection object's detail view and verify.
4. Unrelate the media (from both the media side and the object side). Verify the image disappears from the public browser without editing the media.

No dependencies

Changelog has been updated

@spirosdi ran it locally

no new security vulns


mikejritter (approved):
For some reason the first media I relate isn't showing up. Once I have one media the rest will index as expected. I'm going to spend a little bit of time to debug and see if it's something with my setup because the changes are straightforward.

spirosdi:
> For some reason the first media I relate isn't showing up. Once I have one media the rest will index as expected. I'm going to spend a little bit of time to debug and see if it's something with my setup because the changes are straightforward.

I cannot reproduce the issue actually... I will merge the PR for now. Let us check how it behaves in core.dev too.

PR #555, opened 2026-07-21
DRYD-2131 Search > Qualifiers > 'is/is not' isn't recognizing dates

spirosdi:
**What does this do?**
Changed CoreSessionWrapper from converting NXQL timestamp literals to server-local time to converting them to UTC. Zone-qualified literals are normalized to UTC and unqualified ones are still interpreted as server-local time and then converted to UTC.

https://collectionspace.atlassian.net/browse/DRYD-2131. The new Nuxeo version interprets all timestamp NXQL values as UTC: https://doc.nuxeo.com/nxdoc/nxql/#date-properties. The CoreSessionWrapper was still converting them to server-local time (added in DRYD-114 for older Nuxeo's local-time interpretation). On servers whose JVM runs in a non-UTC timezone, this shifted every date query by the UTC offset, so "is/is not" date searches returned no results.

In order to reproduce the issue locally:
1. Build and deploy current `develop` branch
2. Change in $CATALINA_HOME/bin/setenv.sh `-Duser.timezone=UTC` to `-Duser.timezone=America/Los_Angeles` and then start tomcat
3. Search by an annotation date that should return results and confirm that no results are returned

Testing the fix using staff UI:
1. Build and deploy the feature branch
2. Keep the timezone to `America/Los_Angeles`
3. Search by an annotation date that should return results and confirm that results are returned. 

Testing the fix using different timestamp types:
1. Create a new Object Record using Staff UI with annotation date set to "2026-07-21"
2. The probe() function performs annotation date searches and accepts the timestamp as a parameter:
```
probe() {
  curl -sG -u admin@core.collectionspace.org:Administrator \
    "http://localhost:8180/cspace-services/collectionobjects" \
    --data-urlencode "as=(collectionobjects_annotation:annotationGroupList/*/annotationDate $1)" \
    --data-urlencode "wf_deleted=false" \
    | grep -o 'totalItems>[0-9]*'
}
```
3. probe '= TIMESTAMP "2026-07-21T00:00:00.000Z"'           # expect 1  (what the UI sends UTC)
4. probe '= TIMESTAMP "2026-07-20T17:00:00.000-07:00"'  # expect 1 (offset-qualified)
5. probe '= TIMESTAMP "2026-07-20T17:00:00"'                     # expect 1 (unqualified server-local, DRYD-114 contract preserved)
6. probe '<> TIMESTAMP "2026-07-21T00:00:00.000Z"'        # expect 0 (assuming this is the only annotated record)

No dependencies

No changelog change necessary

@spirosdi ran it locally

no new security vulns

### DIFF
--- a/services/common/src/main/java/org/collectionspace/services/nuxeo/client/java/CoreSessionWrapper.java
+++ b/services/common/src/main/java/org/collectionspace/services/nuxeo/client/java/CoreSessionWrapper.java
@@ -4,10 +4,9 @@ import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
 import java.security.Principal;
-import java.time.Instant;
 import java.time.LocalDateTime;
+import java.time.OffsetDateTime;
 import java.time.ZoneId;
-import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
 import java.util.regex.Matcher;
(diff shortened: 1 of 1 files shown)


mikejritter (approved):
Tested each scenario and verified expected results

PR #556, opened 2026-07-22
DRYD-2117 Procedure/Object > Ability to set Priority Image - top media priority image as search results thumbnail

spirosdi:
**What does this do?**
It includes two changes, to make search-result thumbnails be populated according to the media priority order (along the current way). The changes are:
* In `AdvancedSearch.java`, check if a collectionobject has a media priority list and pick the top entry as the thumbnail in search results. If the list is empty, the endpoint falls back to the existing behavior.
* A new event listener `UpdateMediaPriorityOnDelete`, that removes entries from media priority list when the media record is unrelated or deleted.

This is requested in the scope of the https://collectionspace.atlassian.net/browse/DRYD-2117. The record at the top of the Media priority should be the image shown in the grid and detail-list search result views. The listener is added to avoid showing unrelated/deleted media records as thumbnails. https://collectionspace.atlassian.net/browse/DRYD-2141

1. Create a collectionobject related to two or more media records (with blobs), and set the Media Priority order so the top record is not the alphabetically-first one by title.
2. Search for the object in the staff UI. The grid and detail-list thumbnails should show the top-priority record's image.
3. Search for an object with related media but no priority list. The thumbnail should be unchanged (legacy title-sort behavior).
4. Unrelate the top-priority media record. It should disappear from the object's Media Priority panel, and the thumbnail should shift to the next entry.
5. Delete a top-priority media record. Same result to unrelate.

No dependencies

Changelog has been updated

@spirosdi ran it locally

No new vulns

### DIFF
--- a/3rdparty/nuxeo/nuxeo-platform-listener/updaterelationsondelete/src/main/java/org/collectionspace/services/listener/UpdateMediaPriorityOnDelete.java
+++ b/3rdparty/nuxeo/nuxeo-platform-listener/updaterelationsondelete/src/main/java/org/collectionspace/services/listener/UpdateMediaPriorityOnDelete.java
@@ -7,11 +7,9 @@ import java.util.List;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import org.collectionspace.services.common.api.RefNameUtils;
 import org.collectionspace.services.nuxeo.client.java.CoreSessionInterface;
 import org.collectionspace.services.nuxeo.client.java.CoreSessionWrapper;
 import org.collectionspace.services.nuxeo.listener.AbstractCSEventSyncListenerImpl;
-import org.collectionspace.services.nuxeo.util.NuxeoUtils;
 
 import org.nuxeo.ecm.core.api.DocumentModel;
 import org.nuxeo.ecm.core.api.event.DocumentEventTypes;
@@ -35,6 +33,8 @@ public class UpdateMediaPriorityOnDelete extends AbstractCSEventSyncListenerImpl
     final static String COLLECTIONOBJECTS_COMMON_SCHEMA = "collectionobjects_common";
     final static String SUBJECT_CSID_PROPERTY = "subjectCsid";
     final static String OBJECT_CSID_PROPERTY = "objectCsid";
+    final static String SUBJECT_REFNAME_PROPERTY = "subjectRefName";
+    final static String OBJECT_REFNAME_PROPERTY = "objectRefName";
     final static String SUBJECT_DOCTYPE_PROPERTY = "subjectDocumentType";
     final static String OBJECT_DOCTYPE_PROPERTY = "objectDocumentType";
     final static String MEDIA_PRIORITY_LIST_FIELD = "mediaPriorityList";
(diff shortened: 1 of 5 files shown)


3rdparty/nuxeo/nuxeo-platform-listener/updaterelationsondelete/src/main/java/org/collectionspace/services/listener/UpdateMediaPriorityOnDelete.java
```diff
@@ -0,0 +1,116 @@
...
+        DocumentModel relationDocModel = docContext.getSourceDocument();
+
+        String subjectDocType = (String) relationDocModel.getProperty(RELATIONS_COMMON_SCHEMA, SUBJECT_DOCTYPE_PROPERTY);
+        String objectDocType = (String) relationDocModel.getProperty(RELATIONS_COMMON_SCHEMA, OBJECT_DOCTYPE_PROPERTY);
+
+        String mediaCsid;
+        String collectionObjectCsid;
+
+        if (MEDIA_DOCTYPE.equals(subjectDocType) && COLLECTIONOBJECT_DOCTYPE.equals(objectDocType)) {
+            mediaCsid = (String) relationDocModel.getProperty(RELATIONS_COMMON_SCHEMA, SUBJECT_CSID_PROPERTY);
```
mikejritter: Would it make sense to use the media refname here (`subjectRefName` or `objectRefName`)? Then below we could do `updatedList.remove(mediaRefName)` or `updatedList.removeAll(List.of(mediaRefName))` if we're concerned about possible duplicates.

3rdparty/nuxeo/nuxeo-platform-listener/updaterelationsondelete/src/main/java/org/collectionspace/services/listener/UpdateMediaPriorityOnDelete.java
```diff
@@ -0,0 +1,116 @@
...
+            // The media record was not in the priority list.
+            return;
+        }
+
+        collectionObjectDocModel.setProperty(COLLECTIONOBJECTS_COMMON_SCHEMA, MEDIA_PRIORITY_LIST_FIELD,
+                (Serializable) updatedList);
+        session.saveDocument(collectionObjectDocModel);
+
+        logger.info("Removed media record {} from the media priority list of collectionobject {}",
+                mediaCsid, NuxeoUtils.getCsid(collectionObjectDocModel));
```
mikejritter: We have the CollectionObject csid already so we can reuse it here.

services/advancedsearch/service/src/main/java/org/collectionspace/services/advancedsearch/AdvancedSearch.java
```diff
@@ -151,13 +180,85 @@ private Map<String, String> findBlobInfo(String csid, UriInfoWrapper wrappedUriI
...
+	}
+
+	/**
+	 * Retrieves the blob info (blob CSID and alt text) of a single media record.
+	 *
+	 * @param mediaCsid The CSID of the media record
+	 * @param wrappedUriInfo The wrapped (mutable) UriInfo of the incoming query that ultimately triggered this call
+	 * @return A possibly-empty map of the media record's blob info.
+	 */
+	private Map<String, String> findMediaBlobInfo(String mediaCsid, UriInfoWrapper wrappedUriInfo) {
```
mikejritter, line 223: I was wondering if we could use `mr::get` instead of `getList` in this function. I know it requires the request, resource map, etc, so I'm not sure if it's worth the trouble trying to get everything together for the request vs just querying w/ the advanced search param like we already have. IDK I'll leave it up to you if you think it's worth trying.

mikejritter (commented):
@spirosdi hopefully will be able to test this before the end of the day, if not I'll have to tomorrow morning. I left a few comments just from skimming things over, though everything seems good at a glance.

3rdparty/nuxeo/nuxeo-platform-listener/updaterelationsondelete/src/main/java/org/collectionspace/services/listener/UpdateMediaPriorityOnDelete.java
spirosdi: thanks, done!

3rdparty/nuxeo/nuxeo-platform-listener/updaterelationsondelete/src/main/java/org/collectionspace/services/listener/UpdateMediaPriorityOnDelete.java
spirosdi: done

services/advancedsearch/service/src/main/java/org/collectionspace/services/advancedsearch/AdvancedSearch.java
spirosdi: Yes, basically the get payload parts come back as dom4j rather than W3C DOM so the blob extraction can't be shared and we also wouldn't be able to filter soft-deleted media without an explicit workflow-state check. The list query has the `wf_deleted=false` filtering already. For those, I would rather use `getList`.

mikejritter (approved):
Tested behavior and verified the following:
* Object with multiple related media have deleted csids removed from the media priority list
* When deleting media for an object with a single related media it has no media priority after 

Also see the correct media displaying in the grid and list views.

PR #581, opened 2026-09-02
NO-JIRA: Bound JDBC pool wait with maxWaitMillis so exhaustion fails fast

spirosdi:
**What does this do?**
It adds `maxWaitMillis="5000"` to all five dbcp2 datasources (CspaceDS, NuxeoDS_CS, CsadminDS, Csadmin_NuxeoDS, NuxeoReaderDS) in the context.xml.

The pools are `maxTotal="8"` with no wait limit. So when a pool is exhausted, all upcoming requests are queued without a timeout. The waiting requests accumulate as Tomcat request threads and can exhaust the thread pool (max 200 threads) making Tomcat unavailable. Adding a `maxWaitMillis` limit for queued requests makes them fail early and not accumulate as Tomcat threads. 

Please note that the change converts an unbounded hang into a bounded fast fail. It does not address whatever triggers the pool exhaustion.

Testing site-wide hanging after pool exhaustion:

In order to test, you will need to be able to run bulk concurrent requests. This is not straightforward locally because browsers use HTTP 1 for localhost queries, which has by default a concurrency limit (6). Firefox offers a way to configure this limit. In about:config change the `network.http.max-persistent-connections-per-server` to 100. Also make sure that the `network.http.max-connections` is high enough (like 900).

1. Start local server without the PR change
2. Do a 100:1 bulk relate. You will have to seed your db with at least 100 records. So search for those records, select all of them and click relate. In the dialog, search again for all records and select one of them. 
3. Force pool exhaustion by locking `cspace_default.users` table:
Run the following query to lock:
```
BEGIN;
LOCK TABLE users IN ACCESS EXCLUSIVE MODE;
```
4. After locking `users`, open the network tab in the developer tools and click "Relate selected" in the dialog
5. Watch the requests hang for ever
6. Unlock the `users` table:
```
ROLLBACK;
```

Testing fix after pool exhaustion

1. Start local server with the PR change and repeat the previous steps till step 4
2. Watch that the queued requests (the ones after the pool exhaustion) respond with 500: `Authentication query failed: Cannot get a connection, pool error Timeout waiting for idle object`
3. After unlocking, the ones that made it to the pool resolve with 200

No dependencies

Changelog should be updated

@spirosdi ran it locally

no new security vulnerabilities.


mikejritter (approved):
@spirosdi The only other thing I can think of is for CollectionSpace v8 we might want to add `hibernate.c3p0.checkoutTimeout` as well. Trying variations of your instructions, if I lock `permissions_roles` in the `cspace_default` database then query `accounts/0/accountroles` I can produce the same behavior. I removed c3p0 in 9.0 with a goal of eventually replacing dbcp2 with HikariCP, so we might only need the config change in the ucb_8.3 branch as they will be the only client still on 8.x.

spirosdi:
> @spirosdi The only other thing I can think of is for CollectionSpace v8 we might want to add `hibernate.c3p0.checkoutTimeout` as well. Trying variations of your instructions, if I lock `permissions_roles` in the `cspace_default` database then query `accounts/0/accountroles` I can produce the same behavior. I removed c3p0 in 9.0 with a goal of eventually replacing dbcp2 with HikariCP, so we might only need the config change in the ucb_8.3 branch as they will be the only client still on 8.x.

Right, I pushed this commit to ucb_8.1: https://github.com/cspace-deployment/services/commit/90df05f43b1abc213b5a6fe176f9dedb77c6404b with both bounds.

PR #583, opened 2026-09-15
DRYD-2233: Update Resteasy Configuration

mikejritter:
**What does this do?**
This updates our Resteasy deployment configuration so that we only instantiate our Application class a single time:
* Add resteasy-servlet-initializer dependency
* Prefer javax.ws.rs.core.Application as a deployment strategy
* Change CSpaceResteasyBootstrap from extending ResteasyBootstrap to implementing Feature

Jira: https://collectionspace.atlassian.net/browse/DRYD-2233

Using multiple deployment strategies causes the application to misbehave in subtle ways. So far it hasn't impacted runtime, but when adding Micrometer it was causing an exception to occur when scraping the metric registry.

The preference toward the servlet initializer and `javax.ws.rs.core.Application` comes from looking at the resteasy documentation, as it frequently mentions `This section is pretty much deprecated if you are using a Servlet 3.0 container or higher`. For running the CSpaceResteasyBoostrap post-init tasks, multiple ways of running were tried (e.g. as a ServletContextListener), but the requirement of having the ResteasyDeployment object available (and as such, the CollectionSpaceJaxRsApplication) meant that options were limited. The result of this was using a Feature even though it does not exactly match up with its intended use.

I'm not entirely sure how this should be tested, the easiest might be to add some basic logging into CollectionSpaceJaxRsApplication and test that you can see it, e.g.:
```
diff --git a/services/JaxRsServiceProvider/src/main/java/org/collectionspace/services/jaxrs/CollectionSpaceJaxRsApplication.java b/services/JaxRsServiceProvider/src/main/java/org/collectionspace/services/jaxrs/CollectionSpaceJaxRsApplication.java
index d2f4d9ca0..0eab23403 100644
--- a/services/JaxRsServiceProvider/src/main/java/org/collectionspace/services/jaxrs/CollectionSpaceJaxRsApplication.java
+++ b/services/JaxRsServiceProvider/src/main/java/org/collectionspace/services/jaxrs/CollectionSpaceJaxRsApplication.java
@@ -118,6 +118,7 @@ public class CollectionSpaceJaxRsApplication extends Application implements Reso
     private ServletContext servletContext = null;

     public CollectionSpaceJaxRsApplication() {
+        System.out.println("Initializing CollectionSpaceJaxRsApplication");
         //
         // Instantiate all our JaxRS resources
         //
```

Otherwise, build CollectionSpace and run the integration tests against the server.

Testing should be done if any of the startup tasks fail to ensure the server startup does not continue.

No. I don't know if we have any technical docs on how Resteasy is configured in CollectionSpace.

@mikejritter ran the integration tests locally

n/a


mikejritter:
Tested an exception being thrown from CSpaceRestasyBoostrap and confirmed it will cause the webapp startup to halt as expected.

